PR #20, opened 2021-11-02
Start building IR

iph:
Intermediate Representation is needed to resolve a lot of odd
constraints that don't really belong in the parse phase. Examples:

* IR should figure out the correct names for cdk vs not cdk
* IR should figure out ordering issues, if they arise

Post-IR phases should take IR and expect essentially "just instructions"
to be plopped out from the IR in the language of choice. You should be
able to analyze IR and always know what will be emitted, which is also
helpful for tests.

Conditions is the first move to IR, though a lot of it is just duplicate
traversal of the parse tree, which is unfortunate.

### DIFF
--- a/src/ir/conditions.rs
+++ b/src/ir/conditions.rs
@@ -1,0 +1,154 @@
+use crate::parser::condition::{determine_order, ConditionValue};
+use crate::CloudformationParseTree;
+
+// ConditionInstructions are simple assignment + boolean
+// clauses, as conditions are based on those composite values.
+// It may have made more sense to copy completely to the parse tree
+// but for now we will keep ConditionInstruction + ConditionIr
+// as a single entity.
+#[derive(Debug, Clone)]
+pub struct ConditionInstruction {
+    name: String,
+    value: ConditionIr,
+}
+
+#[derive(Debug, Clone)]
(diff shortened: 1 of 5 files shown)


src/ir/conditions.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub enum PseudoParameter {
+    Partition,
+    Region,
+    StackId,
+    StackName,
+    URLSuffix,
+}
```
roblaks, line 154: Should these really live here?

src/ir/conditions.rs
```diff
@@ -0,0 +1,154 @@
...
+            let x = translate_ir(x, parse_tree);
+            ConditionIr::Not(Box::new(x))
+        }
+        ConditionValue::Or(x) => {
+            let or_list = x
+                .iter()
+                .cloned()
+                .map(|y| translate_ir(&y, parse_tree))
+                .collect();
+            ConditionIr::And(or_list)
```
roblaks, line 71: This looks wrong

src/ir/conditions.rs
iph: nah, will move references out

PR #24, opened 2021-11-09
Move all but resources to ir output

iph:
Semantics as an idea has been trasnformed to ir. All of the weird
off-by-one functionality that was there is now all in ir as a single
unit.

IR is now the "I see a parse tree and can make a lot of decisions
from it". It does the semantic-level business logic of combining
multiple areas of the parse tree to make a sensible, eventual output.

Most of parser was also carved out for formatting and output. That was
put into synthesizer instead. This feels more like a compiler now
with specific, and separate responsibilities and phases.


src/synthesizer/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+use crate::ir::CloudformationProgramIr;
+use crate::parser::lookup_table::MappingInnerValue;
+use crate::parser::resource::ResourceValue;
+use crate::parser::sub::{sub_parse_tree, SubValue};
+use crate::CloudformationParseTree;
+use std::collections::HashMap;
+use voca_rs::case::camel_case;
+
+// TODO - remove parse_tree
+pub fn synthesize(parse_tree: CloudformationParseTree, ir: CloudformationProgramIr) {
```
roblaks, line 14: Can we rename this file to typescript_synthesizer?

src/synthesizer/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+        println!(
+            "import * as {} from '{}';",
+            import.name,
+            import.path.join("/")
+        )
+    }
+
+    println!("export interface NoctStackProps extends cdk.StackProps {{");
+    for param in ir.constructor.inputs {
+        println!(
```
roblaks, line 25: For later, but wondering if we should have some type of indent helper that tracks how many indents to put at the start of a line.

roblaks (approved):

src/synthesizer/mod.rs
iph: that would be nice. As of right now i just run programs like:

`cargo run example.json | tsfmt --stdin > example.ts`

Which works just as well. Will think about this in future iterations though.

src/synthesizer/mod.rs
iph: Sure!

roblaks (approved):

PR #25, opened 2021-11-10
Translate ResourceValue to ResourceIr to add metadata

iph:
Resources in intermediate representation was the last holdover
to a working program without manual manipulation of typescript.

This change adds "Complexity", a concept to help translate CDK
interfaces appropriately for L1s, and adds this decoration to
ResourceIr. This change also adds a few other bells and whistles
along the way while debugging final output (like AWS::AccountId)

### DIFF
--- a/src/ir/resources.rs
+++ b/src/ir/resources.rs
@@ -1,1 +1,297 @@
+use crate::ir::reference::{Origin, Reference};
+use crate::parser::resource::ResourceValue;
+use crate::parser::sub::{sub_parse_tree, SubValue};
+use crate::specification::{spec, Complexity, Specification};
+use crate::{CloudformationParseTree, TransmuteError};
+use std::collections::HashMap;
 
+// ResourceIr is the intermediate representation of a nested stack resource.
+// It is slightly more refined than the ResourceValue, in some cases always resolving
+// known types. It also decorates objects with the necessary information for a separate
+// system to output all the necessary internal structures appropriately.
+#[derive(Clone)]
+pub enum ResourceIr {
+    Null,
+    Bool(bool),
+    Number(i64),
+    String(String),
+
+    // Higher level resolutions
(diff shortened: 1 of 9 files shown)


src/ir/resources.rs
```diff
@@ -1 +1,297 @@
...
+    pub condition: Option<String>,
+    pub resource_type: String,
+    pub properties: HashMap<String, ResourceIr>,
+}
+
+pub fn translates_resources(parse_tree: &CloudformationParseTree) -> Vec<ResourceInstruction> {
+    let spec = spec();
+    let mut resource_instructions = Vec::new();
+    for resource in parse_tree.resources.resources.iter() {
+        let resource_spec = spec
```
roblaks, line 56: Nit: This is more of an optimization question. Does this need to be done for every iteration? Can we save resource_spec after the first time we see the type?

src/ir/resources.rs
```diff
@@ -1 +1,297 @@
...
+                            Complexity::Simple(_) => Option::None,
+                            Complexity::Complex(x) => {
+                                let mut ret =
+                                    format!("{}.{}", resource_translator.resource_type, x);
+                                if x == "Tag" {
+                                    ret = "Tag".to_string();
+                                }
+                                Option::Some(ret)
+                            }
+                        };
```
roblaks, line 152: Worth moving to its own method since it's re-used in two places?

roblaks (approved):

src/ir/resources.rs
iph: I'm a bit confused, as each resource is unique, and the lookup is O(1) already (as it's backed by a hashmap in spec).

src/ir/resources.rs
iph: Yeah. I'ma push so we got a working application but will then update to clean this up.

PR #49, opened 2022-03-17
Add option to write output to a file and fix minor TypeScript coding style & formatting issues.

yilong-wang:
Tested using my own CloudFormation template. The input file is so huge that it is painful to copy the output (which is also huge) from command line. So having the option to write the output to a file is nice.

I also fixed some minor TypeScript coding style & formatting issues after inspecting the output using eslint.

### DIFF
--- a/src/synthesizer/typescript_synthesizer.rs
+++ b/src/synthesizer/typescript_synthesizer.rs
@@ -12,46 +12,63 @@ pub struct TypescriptSynthesizer {
 }
 
 impl TypescriptSynthesizer {
-    pub fn output(ir: CloudformationProgramIr) {
+    pub fn output(ir: CloudformationProgramIr) -> String {
+        let output = &mut String::new();
+
         for import in ir.imports {
-            println!(
+            append_str_with_linebreak(output, &format!(
                 "import * as {} from '{}';",
                 import.name,
                 import.path.join("/")
-            )
+            ));
         }
         // Static imports with base assumptions (e.g. using base 64)
-        println!("import {{Buffer}} from 'buffer';");
+        append_str_with_linebreak(output, "import {Buffer} from 'buffer';");
+        append_str_with_linebreak(output, "\n// Interfaces");
+        append_str_with_linebreak(output, "export interface NoctStackProps extends cdk.StackProps {");
 
-        println!("export interface NoctStackProps extends cdk.StackProps {{");
         for param in ir.constructor.inputs {
-            println!(
-                "\treadonly {}: {}",
+            append_str_with_linebreak(output, &format!(
+                "\treadonly {}: {};",
                 camel_case(&param.name),
                 camel_case(&param.constructor_type)
-            )
+            ));
         }
-        println!("}}");
-        println!("export class NoctStack extends cdk.Stack {{");
-        println!("\tconstructor(scope: cdk.App, id: string, props: NoctStackProps){{");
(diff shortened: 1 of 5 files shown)


src/synthesizer/typescript_synthesizer.rs
```diff
@@ -12,155 +12,193 @@ pub struct TypescriptSynthesizer {
...
                     camel_case(&reference.name),
-                    deletion_policy
-                );
+                    deletion_policy,
+                ));
             }
+
             if !reference.dependencies.is_empty() {
-                println!("{}.addOverride('DependsOn', [", camel_case(&reference.name));
+                append_str_with_linebreak(output, &format!("{}.addOverride('DependsOn', [", camel_case(&reference.name)));
```
iph, line 146: For brevity, would `line_break` be better? Since the first one is `output` gives away the goal.

src/synthesizer/typescript_synthesizer.rs
```diff
@@ -229,13 +267,17 @@ pub fn to_string_ir(resource_value: &ResourceIr) -> Option<String> {
...
             };
 
-            Option::Some(format!("({})?{}:{}", bool_expr, true_expr, false_expr))
+            if false_expr == "{}" || false_expr == "[]" {
+                // false_expr here ("{}" or "[]") might give a type mismatch error so we remove it.
+                Option::Some(format!("{}", true_expr))
+            } else {
+                Option::Some(format!("{} ? {} : {}", bool_expr, true_expr, false_expr))
+            }
         }
```
iph, line 281: The problem is that actually changes semantics of the if statement -- and we can't compress to only the true value

As am example:

```
{ "Fn::If" ["OnDemand", { "ProvisionedWriteThroughput": 10 }, "AWS::NoValue"] }
 ```

No Value turns into a false_expr of `{}` -- which is technically incorrect as well -- as I learned through experimentation 😂. but translating that to only the true expression would actually be an incorrect program and lead to an outage in certain scenarios (imagine instead of swapping to null above, it just always output ProvisionedWriteThroughput).

In this case, I think we should swap false_expr to return `undefined`. Props would already have a union of undefined in cases where a user can optionally supply values. This preserves the original semantics, and will lead to correct TS programs -- opposed to `{}` which is an `Any` type -- which would not satisfy interfaces.

iph: Also, for the [] -- I think in all cases an empty array would satisfy any type that required an array so we can just keep that one.

iph (commented):
Overall a great addition -- just some verbosity to remove and we need the if statement (explained below)

yilong-wang:
LMAO I might have done something wrong that caused this PR being closed.

src/synthesizer/typescript_synthesizer.rs
iph: I honestly think that means the if statement is unreachable (as in if the if statement went on the false path, the cfn template would fail). Sadly, noctilucent can't detect that as it's a semantic error.

yilong-wang:
I am so confused now. Why is this PR not updated after I pushed a newer commit - https://github.com/yilong-wang/noctilucent/commits/main.

It is supposed to work according to https://stackoverflow.com/questions/9790448/how-to-update-a-pull-request-from-forked-repo.

yilong-wang:
Alright, I created a new PR https://github.com/iph/noctilucent/pull/50. Will close this one.

PR #50, opened 2022-03-17
Add option to write output to a file and fix minor TypeScript coding style & formatting issues.

yilong-wang:
[Previous PR](https://github.com/iph/noctilucent/pull/49) cannot be updated for some reason. So I am creating a new one. 

**Can you help squash the commits on the UI before merging?**

### DIFF
--- a/src/synthesizer/typescript_synthesizer.rs
+++ b/src/synthesizer/typescript_synthesizer.rs
@@ -115,11 +115,10 @@ impl TypescriptSynthesizer {
                                 "{}: {}{}",
                                 camel_case(name),
                                 x,
-                                // Remove trailing comma.
-                                if i == reference.properties.len() - 1 {
-                                    ""
-                                } else {
-                                    ","
+                                match i {
+                                    // Remove trailing comma.
+                                    x if x == reference.properties.len() - 1 => "",
+                                    _ => ",",
                                 }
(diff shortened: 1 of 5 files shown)


src/synthesizer/typescript_synthesizer.rs
```diff
@@ -229,13 +305,24 @@ pub fn to_string_ir(resource_value: &ResourceIr) -> Option<String> {
...
+                bool_expr,
+                true_expr,
+                // Convert "{}" to undefined to avoid type mismatch errors. This works for most cases
+                // but not all, e.g., Type 'undefined' is not assignable to type 'IResolvable | PolicyProperty'.
+                // As of now, the user should manually fix when still seeing type mismatch errors.
+                if false_expr == "{}" {
+                    "undefined".to_string()
+                } else {
+                    false_expr
+                }
```
iph, line 324: iirc: It's more idiomatic to use `match`. Also, I _think_ the only time `{}` comes about is above in the `None` case -- so we can just replace to undefined there.

yilong-wang: Thanks. That is indeed cleaner.

yilong-wang:
Sent out new revision.

iph (approved):

PR #63, opened 2022-09-03
Add capability to parse and convert floats in CFN templates

vtanguturi:
**What is the problem?**

As is noctilucent is unable to convert a CFN template if there exists floats in the template. I ran into this on the Threshold property for a CloudWatch alarm, but looking at the specification, there are a couple different resources that could run into this (any property that allows a Double)

**What is the solution?**

While parsing, we will check if a string is an integer or not, using https://docs.rs/numberkit/latest/numberkit/fn.is_digit.html. If it is we will continue with today's parsing, otherwise we will use a new ResourceValue::Double. The not straightforward thing that I had to do was create essentially a wrapper around the f64 type as it does not implement Eq. 

**Example error**

when trying to convert the following resource:

```
    "AlarmResource": {
      "DependsOn": "OtherResource",
      "Properties": {
        "AlarmDescription": "blah",
        "AlarmName": "blah",
        "ComparisonOperator": "GreaterThanOrEqualToThreshold",
        "EvaluationPeriods": 3,
        (... Other Metadata ...)
        "Threshold": 3.5,
        "TreatMissingData": "breaching"
      },
      "Type": "AWS::CloudWatch::Alarm"
    },
```

I get the following error:

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/parser/resource.rs:152:72
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:143:14
   2: core::panicking::panic
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:48:5
   3: noctilucent::parser::resource::build_resources_recursively
   4: noctilucent::parser::resource::build_resources
   5: noctilucent::CloudformationParseTree::build
   6: noctilucent::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

**After this change**

I am able to get the stack converted 💯 
(Generated typescript:

```
const alarmResource = new cloudwatch.CfnAlarm(this, 'AlarmResource', {
evaluationPeriods: 3,
threshold: 3.5,
alarmDescription: 'blah',
comparisonOperator: 'GreaterThanOrEqualToThreshold',
treatMissingData: 'breaching',
alarmName: `blah`
});
alarmResource.addOverride('DependsOn', ['OtherResource']);
```

(Disclaimer: I've never written rust before so let me know if there is a better way to do this)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ edition = "2018"
 clap = "3.0.0-beta.2"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+numberkit = "0.1.0"
 nom = "7.0.0"
 voca_rs = "1.14.0"
 topological-sort = "0.1.0"
--- a/src/ir/resources.rs
+++ b/src/ir/resources.rs
@@ -1,5 +1,5 @@
 use crate::ir::reference::{Origin, Reference};
-use crate::parser::resource::ResourceValue;
+use crate::parser::resource::{ResourceValue, WrapperF64};
 use crate::parser::sub::{sub_parse_tree, SubValue};
 use crate::specification::{spec, Complexity, SimpleType, Specification};
 use crate::{CloudformationParseTree, TransmuteError};
@@ -16,6 +16,7 @@ pub enum ResourceIr {
     Null,
     Bool(bool),
     Number(i64),
+    Double(WrapperF64),
     String(String),
 
     // Higher level resolutions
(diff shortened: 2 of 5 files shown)


iph (approved):
it's beautiful :')

PR #102, opened 2023-04-17
chore: remove dependency on `serde_json`

RomainMuller:
Since all valid JSON is also valid YAML, the `serde_yaml` library can be used to parse all kinds of input, without requiring user selection. Also changes the initial internal representation to be a YAML value instead of a JSON value, which is a little more expressive (although it's not clear whether this is a good thing in this particular case; but at least it reduces the dependency surface).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,9 +11,8 @@ license-file = "LICENSE"
 [dependencies]
 clap = "3.0.0-beta.2"
 serde = { version = "1.0", features = ["derive"] }
-serde_json = "1.0"
 serde_yaml = "0.9"
-serde_with = {version = "2.0.0", features = ["json"]}
+serde_with = { version = "2.0.0", features = ["json"] }
 numberkit = "0.1.0"
 nom = "7.0.0"
 voca_rs = "1.14.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ license-file = "LICENSE"
 clap = "3.0.0-beta.2"
 serde = { version = "1.0", features = ["derive"] }
 serde_yaml = "0.9"
-serde_with = { version = "2.0.0", features = ["json"] }
+serde_with = "2.0.0"
 numberkit = "0.1.0"
 nom = "7.0.0"
 voca_rs = "1.14.0"
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,7 +3,7 @@ use noctilucent::ir::CloudformationProgramIr;
 use noctilucent::synthesizer::typescript_synthesizer::TypescriptSynthesizer;
 use noctilucent::synthesizer::Synthesizer;
 use noctilucent::CloudformationParseTree;
-use serde_json::Value;
+use serde_yaml::Value;
 use std::{fs, io};
 
 fn main() {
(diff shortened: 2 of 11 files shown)


Cargo.toml
```diff
@@ -11,10 +11,9 @@ license-file = "LICENSE"
 [dependencies]
 clap = "3.0.0-beta.2"
 serde = { version = "1.0", features = ["derive"] }
-serde_json = "1.0"
 serde_yaml = "0.9"
-serde_with = {version = "2.0.0", features = ["json"]}
+serde_with = { version = "2.0.0", features = ["json"] }
```
iph, line 15: would we still need json here?

src/main.rs
```diff
@@ -29,19 +29,19 @@ fn main() {
...
                 .required(false)
                 .value_parser(["json", "yaml"])
-                .default_value("json"),
+                .hide(true),
         )
         .get_matches();
 
+    if matches.is_present("inputFormat") {
+        eprintln!("--inputFormat (-f) is a no-op and will be removed in a future version. All input is treated as YAML");
+        eprintln!("as it is a strict super-set of JSON (all valid JSON is valid YAML).");
```
iph, line 38: Note: YAML was added last month, and isn't actually published in a new release. Not sure we need the deprecation notice, but it is a good exercise.

src/parser/condition.rs
```diff
@@ -173,7 +183,7 @@ fn build_condition_recursively(name: &str, obj: &Value) -> Result<ConditionValue
 
                 ConditionValue::Or(v)
             }
-            "!Condition" | "Condition" => {
+            Some("Condition") => {
```
iph, line 186: The sample here says that `!Condition` is real:

```yaml
  CreateBucketPolicy: !And 
    - !Condition IsProduction
    - !Condition CreateBucket
```

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/conditions-section-structure.html#conditions-section-structure-examples

src/parser/resource.rs
```diff
@@ -470,7 +477,7 @@ pub fn build_resources_recursively(
                         Box::new(third_obj),
                     )
                 }
-                "!Ref" | "Ref" => {
+                Some("Ref") => {
```
iph, line 480: I believe !Ref is legit as well: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-ref.html#ref-example

src/parser/condition.rs
```diff
@@ -184,7 +194,7 @@ fn build_condition_recursively(name: &str, obj: &Value) -> Result<ConditionValue
                 };
                 ConditionValue::Condition(condition_name.to_string())
             }
-            "!Ref" | "Ref" => {
+            Some("Ref") => {
```
iph, line 197: Same for Ref here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-ref.html#ref-example

iph (changes requested):
Most of my stuff are nitpicks, but dunno about removing !Ref/!Condition. Take a look at those fellas.

Cargo.toml
RomainMuller: It seems to me like it was in fact _never_ needed 🤣

src/parser/resource.rs
RomainMuller: Ah yes... I made a boo-boo here... Because I changed approaches in the middle of doing this and since those aren't `Fn::` intrinsics I forgot about them...

iph (approved):

HBobertz:
had to fix some merge conflicts so I messed with the branch a bit. All good now

src/main.rs
HBobertz: I think this works for both json and yaml if I'm not mistaken

HBobertz (approved):

PR #104, opened 2023-04-20
fix: additional tests and bug fixes

HBobertz:
added some more basic tests and fixed a simple bug with deletion policies being discovered properly


RomainMuller (approved):
Love it when extra test coverage pays off :P

PR #107, opened 2023-04-21
feat: code-generate PHF lookup maps for the specification

RomainMuller:
Overhauls the modeling of the specification types to be more `enum`-heavy (this provides better semantics than a set of options that are contextually exclusive), and switch from including the complete JSON document of the specification in the binary to code-generating efficient lookup maps at build time. This has the effect of reducing the release binary size from ~11MiB to ~3.1MiB, which is quite an improvement (70% shrink).


iph:
So, I'm down to merge this for the benefits of a slimmer build. I haven't booted this fella up in vscode/intellij, but when I did similar code gen in the past on a 10mb file, it slogged the machine, as the ide attempted to take in the entire structure.

It made programming that code base so unusable that the change was abandoned. Note: different language (this was in Golang), so I'm willing to give this another go for the external benefits -- but back of mind worried about IntelliJ spitting fire at me 😂.

iph (approved):

RomainMuller:
> So, I'm down to merge this for the benefits of a slimmer build. I haven't booted this fella up in vscode/intellij, but when I did similar code gen in the past on a 10mb file, it slogged the machine, as the ide attempted to take in the entire structure.
> 
> It made programming that code base so unusable that the change was abandoned. Note: different language (this was in Golang), so I'm willing to give this another go for the external benefits -- but back of mind worried about IntelliJ spitting fire at me 😂.

I did not notice much of an issue as long as that file isn't opened in the editor (and even after opening it, VSCode appears to fare just fine on my machine).

If this turned out to be a problem, we could actually break the code-gen into multiple smaller files rather easily to make each file easier on the IDE...

PR #108, opened 2023-04-24
chore: new testing method and template test

HBobertz:
Modified the CloudformationParseTree to be more testable and added a template test


src/lib.rs
```diff
@@ -51,7 +51,7 @@ pub struct CloudformationParseTree {
     pub resources: ResourcesParseTree,
     pub outputs: OutputsParseTree,
 
-    logical_lookup: HashSet<String>,
+    pub logical_lookup: HashSet<String>,
```
RomainMuller, line 54: Is this becoming `pub` just for testing? In that case, maybe `pub(crate)` is a less radical solution (generally speaking, I prefer to keep struct fields hidden unless the struct is "pure data", as otherwise your invariants could be broken by whoever owns the value)

tests/json.rs
```diff
@@ -1,4 +1,5 @@
-/// Shamelessly copied from: https://docs.rs/serde_json/latest/src/serde_json/macros.rs.html
+/// Was previously shamelessly copied from: https://docs.rs/serde_json/latest/src/serde_json/macros.rs.html
+/// Constructs a serde::yaml from a json literal (or yaml literal)
```
RomainMuller, line 2: Technically it only accepts JSON input (although yes, this is technically a subset of YAML 🙃)

src/lib.rs
HBobertz: changed it to pub super, pub crate is too private

PR #119, opened 2023-05-01
chore: Adding a contributing guide

Naumel:
Straightforward change, like the title suggests.


HBobertz (approved):
Love that we're adding a contributing guide.

PR #129, opened 2023-05-10
feat: track reference origin nullability

RomainMuller:
Allows emitting optional-chained references when the origin is conditional, so as to avoid the TypeScript compiler complaining about possibly null dereference.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,22 +1,23 @@
 coverage:
   precision: 1
   round: nearest
+  status:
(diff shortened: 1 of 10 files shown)


RomainMuller:
NOTE - Coverage now is branch-aware, and Coverall reports "fully covered" only, so this is a more accurate view, although perhaps a bit scary

Naumel (approved):

iph (approved):

PR #147, opened 2023-05-31
chore: improved code generation API

RomainMuller:
The previous `CodeSink` API immediately flushed out into the output writer, hence preventing additional insertions into the previous code. However, in order to avoid introducing unnecessary imports without having to traverse the entire tree to decide whether a particular import is necessary or not, lazy addition of new imports is desirable.

This introduces a new `CodeBuffer` struct which aggregates code in memory while keeping track of indentation requirements, so that code can be inserted back into previous sections of the file prior to it being written.

This also has the effect of drastically reducing the amount of branches, as there are far fewer failable calls involved when generating code now.


HBobertz (approved):

src/code/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+            trailing_newline,
+        })
+    }
+
+    /// Writes the content of this `CodeBuffer` into the provided writer.
+    pub fn write(self, writer: &mut dyn io::Write) -> io::Result<()> {
+        self.inner_write(&mut IndentedWriter::new(writer))
+    }
+
+    fn inner_write(&self, writer: &mut IndentedWriter) -> io::Result<()> {
```
RomainMuller, line 92: In case you're wondering... I'd have preferred `inner_write` to also consume `self`, however this is not possible because we have to shove the sub-`CodeBuffer`s into `Rc` (reference counting smart pointers), which does not necessarily allow us to actually regain ownership of the underlying value (`Rc::try_unwrap`, as the name implies, only transfers ownership if the reference count is exactly `1`, which may not be the case). This is (in parts) why there is an `inner_write` API that is private here.

PR #167, opened 2023-06-29
chore: add python language target

HBobertz:


tests/end-to-end/simple/app.py
```diff
@@ -0,0 +1,131 @@
...
+    self.bucket_arn = bucket.attrarn if is_us_east1 else None
+    if (is_us_east1):
+      cdk.CfnOutput(self, 'BucketArn', 
+        description = 'The ARN of the bucket in this template!',
+        export_name = 'ExportName',
+        value = self.bucket_arn,
+      )
+
+    self.queue_arn = queue.ref
+    self.is_large = True if is_large_region else False
```
TheRealAmazonKendra, line 129: Sean and I synced just now on this and these should also be `CfnOutput` the difference here is that we don't want to include `export_name` on these two. We also don't want to assign these as class fields because CDK logic will then turn them into exports.

src/synthesizer/python/mod.rs
```diff
@@ -0,0 +1,645 @@
...
+use std::io;
+use std::rc::Rc;
+use voca_rs::case::camel_case;
+
+use super::Synthesizer;
+
+const INDENT: Cow<'static, str> = Cow::Borrowed("  ");
+
+pub struct Python {
+    // TODO: Put options in here for different outputs in Python
```
TheRealAmazonKendra, line 21: Can you clarify what you mean by this TODO?

src/synthesizer/python/mod.rs
```diff
@@ -0,0 +1,645 @@
...
+}
+
+impl Python {
+    #[cfg_attr(coverage_nightly, no_coverage)]
+    #[deprecated(note = "Prefer using the Synthesizer API instead")]
```
TheRealAmazonKendra, line 26: Since we don't consider this module stable yet, we should probably just delete old functionality rather than deprecating, especially since all of this functionality is brand new. Why add an immediately deprecated function?

src/synthesizer/python/mod.rs
```diff
@@ -0,0 +1,645 @@
...
+        ResourceIr::Ref(reference) => output.text(reference.to_python()),
+    }
+    if let Some(trailer) = trailer {
+        output.text(trailer.to_owned())
+    }
+}
+
+fn append_references(output: &CodeBuffer, reference: &ResourceInstruction) {
+    for dep in &reference.references {
+        output.line(format!("if ({dep} is None): raise Exception(\"A combination of conditions caused '{dep}' to be None. Fixit.\")", dep=camel_case(dep)));
```
TheRealAmazonKendra, line 635: This error message feels a little clunky. Can we make it more useful to the user?

tests/end-to-end/simple/app.py
```diff
@@ -0,0 +1,131 @@
...
+              'value': cdk.Fn.base64(table['Values']['String']) if is_us_east1 else _buffer.from('8CiMvAo=', 'base64').toString('binary'),
+            },
+          ],
+        ) if is_us_east1 else None
+    if (bucket is not None):
+      bucket.cfnOptions.metadata = {
+        CostCenter: 1337,
+      }
+      bucket.cfnOptions.deletionPolicy = cdk.CfnDeletionPolicy.RETAIN
+      bucket.addDependency(queue)
```
TheRealAmazonKendra, line 117: Wouldn't these two items need to go inside the if statement?

TheRealAmazonKendra (changes requested):

src/synthesizer/python/mod.rs
HBobertz: Unfortunately I can't. I copied as much of the Typescript synthesizer structure as I possibly could, and that meant needing this Python struct with nothing in it that we just impl. Romain had that comment so I added it too lol

tests/end-to-end/simple/app.py
HBobertz: I think they are, aren't they?

src/synthesizer/python/mod.rs
HBobertz: went ahead and just deleted it.

src/synthesizer/python/mod.rs
HBobertz: Yeah just another thing I did to keep parity with the ts synthesizer. I went ahead and deleted it

src/synthesizer/python/mod.rs
TheRealAmazonKendra: Even if we can't provide more information the message, as it stands, seems pretty idiomatic to Rust, not to the target language so we should edit it to reflect python conventions instead of Rust. Also, `Fixit` probably isn't the wording we want there.

tests/end-to-end/simple/app.py
TheRealAmazonKendra: Oh, you're right. I misread the bracket here.

src/synthesizer/python/mod.rs
TheRealAmazonKendra: Granted, I don't actually use Python so maybe I'm wrong here. I just saw that same-ish style in TypeScript as well.

HBobertz: for context, there was a conversation offline and we agreed that this actually serves 0 practical purpose so we are removing it.

HBobertz:
found a bug with how I'm handling custom stack props while updating the formatting. Gonna leave this broken and will find the time to fix that

HBobertz:
> found a bug with how I'm handling custom stack props while updating the formatting. Gonna leave this broken and will find the time to fix that

Fixed it

TheRealAmazonKendra (approved):

PR #169, opened 2023-06-30
chore: Update to template.yml

Naumel:
Adding CIDR


Naumel:
There is completeness between the simple and VPC test coverage.

PR #175, opened 2023-07-06
fix: issues on cdk diff and changeset in Typescript conversion

wongni:
### Prop names containing special characters in policy documents and resource metadata
* Example cdk diff output
```
kms:EncryptionContext:aws:s3Arn: [
  ...
]
```
* Solution: Wrap the property name with quotes ('')* 

### Prop name case differences in policy documents and resource metadata
* Example cdk diff output
```
[~] AWS::IAM::Role CustomResourceDeadLetterLambdaRole CustomResourceDeadLetterLambdaRole
 ├─ [~] AssumeRolePolicyDocument
  │       ├─ [-] Removed: .Statement
  │       ├─ [-] Removed: .Version
  │       ├─ [+] Added: .statement
  │       └─ [+] Added: .version
```
* Solution: Keep the property names for `policyDocument` and `assumerolepolicydocument`

### CfnOutput's ExportName with Object
* Example code
```
      new cdk.CfnOutput(this, 'MyOutput', {
        exportName: {
          name: 'MyOutput',
        },
        ...
      }
```
* Solution: Unwrap the name value from the object 
```
      new cdk.CfnOutput(this, 'MyOutput', {
        exportName: 'MyOutput',
        ...
      }
```

### CfnOutput's Value with string|undefined type
```
      new cdk.CfnOutput(this, 'MyOutput', {
        value: this.a2zDnsHostZone,
        ...
      }
```
* Solution: Add ! at the end
```
      new cdk.CfnOutput(this, 'MyOutput', {
        value: this.a2zDnsHostZone!,
        ...
      }
```

### DIFF
--- a/src/synthesizer/typescript/mod.rs
+++ b/src/synthesizer/typescript/mod.rs
@@ -189,11 +189,11 @@ impl Synthesizer for Typescript {
                     ));
                     ctor.text(format!("{INDENT}? "));
                     let indented = ctor.indent(INDENT);
-                    emit_resource_ir(context, &indented, &op.value, Some("\n"));
+                    emit_resource_ir(context, &indented, &op.value, Some("\n"), false);
                     ctor.line(format!("{INDENT}: undefined;"));
                 } else {
                     ctor.text(format!("this.{var_name} = "));
-                    emit_resource_ir(context, &ctor, &op.value, Some(";\n"));
+                    emit_resource_ir(context, &ctor, &op.value, Some(";\n"), false);
                 }
 
                 if let Some(export) = &op.export {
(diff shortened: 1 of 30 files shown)


src/synthesizer/typescript/mod.rs
```diff
@@ -394,8 +413,14 @@ fn emit_resource_metadata(
     match metadata {
         ResourceIr::Object(_, entries) => {
             for (name, value) in entries {
-                output.text(format!("{name}: "));
-                emit_resource_ir(context, &output, value, Some(",\n"));
+                // Wrap the prop name with quotes ('') to support special characters in it
+                output.text(format!("'{name}': "));
```
iph, line 417: I personally preferred the non quoted version (kept going back and forth on the initial implementation):\ 

Do we know the characters that aren't allowed, and we can escape them (or is that too much logic?) In the end, I want folks to not have to rewrite a bunch, and I feel most would remove the quotes.

wongni: @iph I couldn't find what chars are allowed or not allowed in the Javascript properties without quotes. This (https://mathiasbynens.be/notes/javascript-properties) doc might help.

iph: fair enough. I think it's not worth complicating -- so I'm cool with this. Running workflows

PR #185, opened 2023-07-10
fix(all): properly transform depends_on, and selection

iph:
With the move to proper typing in parsing, we accidentally shed some valuable transforms that CloudFormation takes on behalf of the customer. This brings back those transforms.

Specifically:
- DependsOn is allowed to have a single value
- Select index can be a string that is parse-able into an int


HBobertz (approved):
awesome thanks

PR #188, opened 2023-07-13
feat: Java support

Naumel:
Documented the edge-case of typed resource properties in `CDK Migrate - Burndown`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,6 @@ serde_yaml = "^0.9.22"
 topological-sort = "^0.2.2"
 voca_rs = "^1.15.2"
 wasm-bindgen = "^0.2.87"
-serde_json = "^1.0.97"
 
 
 [build-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ homepage = "https://github.com/iph/noctilucent#readme"
 default = ["typescript"]
 typescript = []
 golang = []
+java = []
 
 [lib]
 crate-type = ["cdylib", "lib"]
@@ -39,6 +40,7 @@ serde_yaml = "^0.9.22"
 topological-sort = "^0.2.2"
 voca_rs = "^1.15.2"
 wasm-bindgen = "^0.2.87"
+serde_json = "^1.0.97"
 
 
 [build-dependencies]
--- a/src/synthesizer/java/mod.rs
+++ b/src/synthesizer/java/mod.rs
@@ -9,7 +9,7 @@ use crate::parser::lookup_table::MappingInnerValue;
 use crate::specification::{CfnType, Structure};
 use std::any::{type_name};
 use std::borrow::Cow;
-use std::fmt::{Debug};
+use std::fmt::{Debug, format};
 use std::io;
 use std::rc::Rc;
 use super::Synthesizer;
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -39,6 +40,7 @@ serde_yaml = "^0.9.22"
 topological-sort = "^0.2.2"
 voca_rs = "^1.15.2"
 wasm-bindgen = "^0.2.87"
+serde_json = "^1.0.97"
```
RomainMuller, line 43: Seems redundant, I can't find any use of that...

src/synthesizer/java/mod.rs
```diff
@@ -0,0 +1,649 @@
+#![allow(unused_variables)]
+
```
RomainMuller, line 2: Probably wanna drop this now.

src/synthesizer/java/mod.rs
```diff
@@ -0,0 +1,649 @@
...
+            $(class.line(format!($lines));)*
+        }
+    };
+}
+
+fn debug<T: Debug>(value: T, prefix: &str) {
+    let type_name = type_name::<T>();
+    println!("[{:?}]: {:?}", prefix, value);
+}
```
RomainMuller, line 41: Apparently this is dead code :)

tests/end-to-end/simple/App.java
```diff
@@ -0,0 +1,185 @@
...
+        final Map<String, T> map = inner.getOrDefault(primaryKey, new TreeMap<>());
+        map.put(secondaryKey, value);
+        inner.put(primaryKey, map);
+        return this;
+    }
+
+    public CfnMapping get() {
+        return CfnMapping.Builder.create(this.scope, this.name).mapping(this.inner).build();
+    }
+}
```
RomainMuller, line 69: This probably should be moved _after_ the actual business the customer is here for (that is, the app & stack).

tests/end-to-end/simple/App.java
```diff
@@ -0,0 +1,185 @@
...
+
+public class NoctApp {
+	public static void main(final String[] args) {
```
RomainMuller, line 72: This is odd/wonky - looks like this is intended with `\t` while the previous code is indented with `    `... Should use the same indent everywhere.

And then further down there is some indentation oddities...

src/synthesizer/java/mod.rs
```diff
@@ -0,0 +1,649 @@
...
+use std::borrow::Cow;
+use std::fmt::{Debug};
+use std::io;
+use std::rc::Rc;
+use super::Synthesizer;
+use voca_rs::case::camel_case;
+use crate::ir::outputs::OutputInstruction;
+use crate::ir::reference::{Origin, PseudoParameter, Reference};
+
+const INDENT: Cow<'static, str> = Cow::Borrowed("\t");
```
RomainMuller, line 20: The Google style guide for Java is fairly popular and demands two spaces is used for intent.

```suggestion
const INDENT: Cow<'static, str> = Cow::Borrowed("  ");
```

tests/end-to-end/simple/App.java
```diff
@@ -0,0 +1,185 @@
...
+}
+interface NoctStackProps extends StackProps {
+}
+class NoctStack extends Stack {
+  public NoctStack(final Construct scope, final String id) {
+    super(scope, id, null);
+  }
+  public NoctStack(final Construct scope, final String id, final StackProps props) {
+    super(scope, id, props);
+    { // Start Mapping section
```
RomainMuller, line 38: If we wrap the mappings section in `{}` I guess none of the bindings it makes are visible in the rest of the code? Maybe should remove the braces there & optionally just keep a leading comment?

tests/end-to-end/simple/App.java
```diff
@@ -0,0 +1,185 @@
+package com.acme.test.simple;
+
+
```
RomainMuller, line 3: Is the duplicated blank line intentional here?

RomainMuller (approved):

PR #194, opened 2023-07-17
feat: enable users to define stack name

TheRealAmazonKendra:
This change adds a CLI input `--stack-name` that users can specify. The default value of `NoctStack` is unchanged if no input is provided by the user.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -91,13 +91,10 @@ fn main() -> anyhow::Result<()> {
         unsupported => panic!("unsupported language: {}", unsupported),
     };
 
-    let stack_name: &str = match matches
+    let stack_name = matches
         .get_one::<String>("stack-name")
         .map(String::as_str)
-        .unwrap_or("NoctStack") 
-    {
-        stack_name => stack_name,
-    };
+        .unwrap_or("NoctStack");
 
     ir.synthesize(synthesizer.as_ref(), &mut output, stack_name)?;
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -46,6 +46,14 @@ fn main() -> anyhow::Result<()> {
                 .value_parser(targets)
                 .action(ArgAction::Set),
         )
+        .arg(
+            Arg::new("stack-name")
+                .help("Sets the name of the stack")
+                .required(false)
+                .long("stack-name")
+                .short('s')
+                .action(ArgAction::Set),
+        )
         .get_matches();
 
     let cfn_tree: CloudformationParseTree = {
@@ -83,7 +91,15 @@ fn main() -> anyhow::Result<()> {
         unsupported => panic!("unsupported language: {}", unsupported),
     };
 
-    ir.synthesize(synthesizer.as_ref(), &mut output)?;
+    let stack_name: &str = match matches
+        .get_one::<String>("stack-name")
+        .map(String::as_str)
+        .unwrap_or("NoctStack") 
+    {
+        stack_name => stack_name,
+    };
+
+    ir.synthesize(synthesizer.as_ref(), &mut output, stack_name)?;
 
     Ok(())
 }
--- a/tests/end-to-end/simple/app.go
+++ b/tests/end-to-end/simple/app.go
@@ -16,8 +16,8 @@ type SimpleStackProps struct {
 	BucketNamePrefix *string
 }
 
-// / An example stack that uses many of the syntax elements permitted in a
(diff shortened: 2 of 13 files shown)


TheRealAmazonKendra:
Looks like my IDE added a bunch of spacing where it didn't really need to be. I'll fix that. Any specific settings I should be setting for VSCode?

src/main.rs
```diff
@@ -83,7 +91,15 @@ fn main() -> anyhow::Result<()> {
...
     };
 
-    ir.synthesize(synthesizer.as_ref(), &mut output)?;
+    let stack_name: &str = match matches
+        .get_one::<String>("stack-name")
+        .map(String::as_str)
+        .unwrap_or("NoctStack") 
+    {
+        stack_name => stack_name,
+    };
```
iph, line 100: Most of the other options had the `unwrap_or` as essentially "if null", e.g.
```
    let mut output: Box<dyn io::Write> = match matches
        .get_one::<String>("OUTPUT")
        .map(String::as_str)
        .unwrap_or("-")
    {
        "-" => Box::new(io::stdout()),
        output_file => Box::new(fs::File::create(output_file)?),
    };
```

Since your code space is always doing the same action, you can simplify it to have no matches, so:
```
    let stack_name = matches
        .get_one::<String>("stack-name")
        .map(String::as_str)
        .unwrap_or("NoctStack");
```

(I also remove the `: &str`, as it was unambiguously determined on the right side)

tests/end-to-end/simple/app.go
```diff
@@ -28,42 +28,42 @@ type NoctStack struct {
...
 	/*
-	booleans := map[*string]map[*string]*bool{
-		jsii.String("True"): map[*string]*bool{
-			jsii.String("true"): jsii.Bool(true),
-		},
-		jsii.String("False"): map[*string]*bool{
-			jsii.String("false"): jsii.Bool(false),
-		},
-	}
+		booleans := map[*string]map[*string]*bool{
```
iph, line 33: interesting side effect :o

tests/end-to-end/simple/app.go
```diff
@@ -152,10 +152,10 @@ func NewNoctStack(scope constructs.Construct, id string, props NoctStackProps) *
...
 }
 
-/// ifCondition is a helper function that replicates the ternary
-/// operator that can be found in other languages. It is conceptually
-/// equivalent to writing `cond ? whenTrue : whenFalse`, meaning it
-/// returns `whenTrue` if `cond` is `true`, and `whenFalse` otherwise.
+// / ifCondition is a helper function that replicates the ternary
+// / operator that can be found in other languages. It is conceptually
+// / equivalent to writing `cond ? whenTrue : whenFalse`, meaning it
+// / returns `whenTrue` if `cond` is `true`, and `whenFalse` otherwise.
```
iph, line 158: I think this means it's no longer a specific type of doc comment, so we should undo this one?

iph (commented):
Unsure what system is doing the whitespace changes -- one small nitpick over `matches` but lgtm overall.

iph:
> Looks like my IDE added a bunch of spacing where it didn't really need to be. I'll fix that. Any specific settings I should be setting for VSCode?

So for the Rust code, it's totally fine because at the end you just: `cargo fmt` to get it to the desired end state. For everything in end-to-end, see if you can just `ignore those`, since it's all auto-generated.

TheRealAmazonKendra:
> Unsure what system is doing the whitespace changes -- one small nitpick over `matches` but lgtm overall.

I figured it out. It's my go linter. Turning that off.

src/main.rs
```diff
@@ -83,7 +91,12 @@ fn main() -> anyhow::Result<()> {
         unsupported => panic!("unsupported language: {}", unsupported),
     };
 
-    ir.synthesize(synthesizer.as_ref(), &mut output)?;
+    let stack_name = matches
+        .get_one::<String>("stack-name")
+        .map(String::as_str)
+        .unwrap_or("NoctStack");
```
HBobertz, line 97: totally chill we're doing this like this for now. But at some point we need to incorporate cfn's get-template somewhere around this bit to get the stack name of an already deployed stack

TheRealAmazonKendra: How does that mesh with the input and output being filepaths?

HBobertz: nvm you did it above I just didnt notice. We're all gucci

HBobertz (approved):

PR #209, opened 2023-08-09
feat(java): add resource attributes

TheRealAmazonKendra:
Java was missing Metadata, DeletionPolicy, UpdatePolicy, and DependsOn. This PR adds those.


madeline-k (approved):
This looks good to me. but I am having some difficulty reviewing, because the different synthesizer implementations don't align to each other very well. It might be worth it to refactor and have them more closely aligned/sharing more Traits. I think this may be missing from the Go synthesizer as well.

PR #213, opened 2023-08-16
fix(go, java): synthesizing a go or java stack from a template with CFN properties of type json fails

### LINKED ISSUE #210: go: synthesizing a template with CFN properties of type json fails
a template including this:
```
"MyRDMessageQueueGroup" : {
      "Type" : "AWS::IAM::Group",
      "Properties" : {
        "Policies": [ {
          "PolicyName": "MyQueueGroupPolicy",
          "PolicyDocument": {
            "Statement":[ {
              "Effect":"Allow",
 	      "Action":[ "sqs:DeleteMessage", "sqs:ReceiveMessage" ],
              "Resource":[ { "Fn::GetAtt" : ["MyQueue1", "Arn"]}, { "Fn::GetAtt" : ["MyQueue2", "Arn"]} ]
            } ]
          }
        } ]
      }
    }
```
Will fail to synthesize to go with the error: 
```
thread 'main' panicked at 'internal error: entered unreachable code: object with simple structure (Json)', src/synthesizer/golang/mod.rs:637:29
```

This is because the type of a PolicyDocument is json ([docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-iam-policy-2.html#cfn-iam-policies-policydocument)), and the go synthesizer doesn't currently handle this case https://github.com/iph/noctilucent/blob/e75176e31e4af8b157d28b825ea9f995e4c7bcae/src/synthesizer/golang/mod.rs#L631.

madeline-k:
fixes #210

### DIFF
--- a/src/synthesizer/golang/mod.rs
+++ b/src/synthesizer/golang/mod.rs
@@ -294,6 +294,81 @@ impl Synthesizer for Golang {
             output.value.emit_golang(context, &fields, None);
             fields.line(",");
         }
+        code.newline();
+
+        let main_block = code.indent_with_options(IndentOptions {
+            indent: INDENT,
+            leading: Some("func main() {".into()),
+            trailing: Some("}".into()),
+            trailing_newline: true,
+        });
+
+        main_block.line("defer jsii.Close()");
+        main_block.newline();
+        main_block.line("app := cdk.NewApp(nil)");
+        main_block.newline();
+        let split_stack_name: Vec<&str> = stack_name.split("Stack").collect();
+        main_block.line(format!(
+            "New{stack_name}(app, \"{}\", {stack_name}Props{{",
+            split_stack_name[0]
+        ));
+        main_block.indent(INDENT).line("cdk.StackProps{");
+        main_block.indent(INDENT).indent(INDENT).line("Env: env(),");
+        main_block.indent(INDENT).line("},");
+        for param in &ir.constructor.inputs {
+            if param.default_value.is_some() {
+                main_block.indent(INDENT).line(format!(
+                    "{}: \"{}\",",
+                    golang_identifier(&param.name, IdentifierKind::Exported),
+                    param.default_value.clone().unwrap()
+                ));
+            }
+        }
+        main_block.line("})");
+        main_block.newline();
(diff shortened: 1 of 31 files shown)


src/synthesizer/golang/mod.rs
```diff
@@ -620,25 +620,37 @@ impl GolangEmitter for ResourceIr {
...
+                        props.text(format!(
+                            "\"{name}\": ",
+                            name = golang_identifier(name, IdentifierKind::Exported)
+                        ));
+                    } else {
+                        props.text(format!(
+                            "{name}: ",
+                            name = golang_identifier(name, IdentifierKind::Exported)
+                        ));
+                    }
```
madeline-k, line 653: This is required because the format string for the keys needs to be different depending on if it's for a "real" type or a just a json blob. Json blobs are represented as `map[string]interface{}` in go, which means the keys needs to be wrapped in double quotes `"`.

The way it is written is pretty verbose, and I am new to rust so please let me know if there is a more concise way to write it! I was hoping to do something like the below, 
but this doesn't work because the format string in the `format!` macro MUST be a string literal. There is no way to define it dynamically that I have found. 
```
props.text(format!(
    if structure_is_simple_json { "\"{name}\": " } else { "{name}: " },
    name = golang_identifier(name, IdentifierKind::Exported)
));
```

src/synthesizer/golang/mod.rs
```diff
@@ -620,25 +620,37 @@ impl GolangEmitter for ResourceIr {
...
                             name => format!("&{name}/* FIXME */{{").into(),
                         },
-                        Structure::Simple(cfn) => {
-                            unreachable!("object with simple structure ({:?})", cfn)
+                        Structure::Simple(cfn) => match cfn {
+                            CfnType::Json => {
+                                structure_is_simple_json = true;
+                                "map[string]interface{} {".into()
+                            }
+                            _ => unreachable!("object with simple structure ({:?})", cfn)
```
madeline-k, line 636: Now this case should be truly unreachable, because there should never be an `Object` where its structure is one of the basic types.

HBobertz (approved):

Issue #214, opened 2023-08-17
csharp: support for synthesizing apps in C#


madeline-k:
Status: Work in Progress

## To implement
- [ ] Import statements
- [x] Parameters as construct props
- [x] Resources
- [x] Mappings
- [x] Conditions
- [x] Outputs
- [x] !Ref
- [x] !GetAtt
- [x] !FindInMap
- [x] !Join
- [x] !Sub
- [x] !Base64
- [x] !ImportValue
- [x] !Select
- [x] !Cidr
- [x] !GetAZs
- [ ] Conditional resource creation
    - [ ] DeletionPolicy
    - [ ] UpdatePolicy
    - [ ] Metadata
    - [ ] DependsOn
- [x] Description

## Challenges unique to C#
- [ ] The casing for using statements is not consistent. Acronyms are all caps (`Amazon.CDK.IAM`) but names of services are capitalized (`Amazon.CDK.Lambda`). To know which is which, we need to fetch this from somewhere. 
- [x] Defining the type for each input prop
- [x] Defining the type for each output prop
- [x] Type names for composite nested properties need to be explicitly included - i.e. the [WebsiteConfiguration](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-websiteconfiguration.html) property must be declared with `new CfnBucket.WebsiteConfigurationProperty`([C# doc link](https://docs.aws.amazon.com/cdk/api/v2/dotnet/api/Amazon.CDK.AWS.S3.CfnBucket.WebsiteConfigurationProperty.html))

## Optimizations
- [ ] Only create local variables for resources that will be referenced again later
- [ ] Extract shared code snippets to helper functions. (I didn't do this yet, because too much abstraction makes it harder to read and understand. So I'm not sure it's worth it to avoid a little bit of repetition)

PR #227, opened 2023-08-31
feat: add C# support

madeline-k:
See issue #214 for more details.

C# is mostly following the model of Golang, but there are some elements from TypeScript/Python/Java, as well as some unique elements.

I have manually compiled and synthesized the resulting c# code in the end-to-end test files in this PR.


tests/end-to-end/simple/App.cs
```diff
@@ -0,0 +1,195 @@
...
+                            { "false", false },
+                    }
+                },
+            };
+            var lists = new Dictionary<string, Dictionary<string,string[]>> 
+            {
+                {
+                    "Candidates", new Dictionary<string, string[]>
+                    {
+                            {
```
TheRealAmazonKendra, line 63: Is the double indent intentional here? I'm not totally familiar with C# spacing.

madeline-k: Thanks for catching! That is not intentional.

madeline-k: Fixed

tests/end-to-end/simple/App.cs
```diff
@@ -0,0 +1,203 @@
...
+using Constructs;
+using System.Collections.Generic;
+
+namespace Com.Acme.Test.Simple
+{
+    public class SimpleStackProps : StackProps
+    {
+        /// <summary>
+        /// The prefix for the bucket name
+        /// </summary>
```
HBobertz, line 13: I was gonna ask if we could use multi line comments here, but this seems to be the standard which is horrifying https://www.w3schools.com/cs/cs_comments.php#:~:text=C%23%20Multi%2Dline%20Comments,will%20be%20ignored%20by%20C%23.

tests/end-to-end/simple/App.cs
```diff
@@ -0,0 +1,203 @@
...
+        {
+            // Applying default props
+            props.BucketNamePrefix ??= "bucket";
+            props.LogDestinationBucketName = new CfnParameter(this, "LogDestinationBucketName", new CfnParameterProps
+            {
+                Type = "AWS::SSM::Parameter::Value<String>",
+                Default = props.LogDestinationBucketName ?? "/logging/bucket/name",
+            }).ValueAsString;
+
+            var booleans = new Dictionary<string, Dictionary<string,bool>>
```
HBobertz, line 51: For consistency purposes should we have a // Mappings comment above this line

tests/end-to-end/simple/App.cs
```diff
@@ -0,0 +1,203 @@
...
+                                "2",
+                                "3",
+                            }
+                        },
+
+                        { "Number", 42 },
+                        { "String", "Baz" },
+                    }
+                },
+            };
```
HBobertz, line 142: Somewhat nitty but is there 
a) no way to shrink the whitespace in the mappings? It's nearly double the size of other synthesizers
b) Should we have a newline gap here and a comment to indicate the next 3 booleans are the conditions

tests/end-to-end/simple/App.cs
```diff
@@ -0,0 +1,203 @@
...
+            bool isUs = Fn.Select(0, Fn.Split("-", Region)) == "us";
+            bool isUsEast1 = Region == "us-east-1";
+            bool isLargeRegion = isUsEast1;
+            var queue = new CfnQueue(this, "Queue", new CfnQueueProps
```
HBobertz, line 146: Again I personally like the Gap + comment to label this as the resources section.

HBobertz (changes requested):
Just some comment/whitespace stuff. If we want we can just merge and do this as a fast follow

tests/end-to-end/simple/App.cs
madeline-k: Yes I went to a lot of trouble to find examples in the wild to confirm that this is the c# style.

README.md
```diff
@@ -20,6 +20,7 @@ Name         | Enabled by default | Description
 `golang`     | :heavy_check_mark: | Enables support for Go output
 `java`       | :heavy_check_mark: | Enables support for Java output
 `Python`     | :heavy_check_mark: | Enables support for Python output
+`csharp`     |                    | Enables support for C# output
```
TheRealAmazonKendra, line 23:
```suggestion
`csharp`     | :heavy_check_mark: | Enables support for C# output
```

tests/end-to-end/simple/App.cs
TheRealAmazonKendra: The first question here should be if this spacing is the C# convention. If so, it's the spacing we should use.

tests/end-to-end/vpc/App.cs
```diff
@@ -0,0 +1,52 @@
+using Amazon.CDK;
+using Amazon.CDK.AWS.EC2;
+using Constructs;
+using System.Collections.Generic;
+
+namespace Com.Acme.Test.Simple
+{
+    public class VpcStackProps : StackProps
+    {
+    }
```
TheRealAmazonKendra, line 10: Should we omit these if no props are present?

TheRealAmazonKendra (changes requested):

tests/end-to-end/vpc/App.cs
madeline-k: I have the same question. I am on the fence on this one. I decided to leave it in for two reasons:
1. Typescript and Go both do it this way
2. This shows the customer how they can start expanding their CDK app code to be more useful/flexible in a CDK idiomatic way

But, I can also see that it's adding clutter to the generated file.

tests/end-to-end/simple/App.cs
madeline-k: @HBobertz , I addressed b).

For the spacing, good news! I just learned there is a way to do index-based initialization in C#, so that already makes it shorter. It's still long though so let me lay out the options. in C#, there are two ways to format your code block in curly braces:
all in one line
```csharp
new Thing { Param1 = 1, Param2 = 2 }
```

OR taking up at least 4 lines (or it would be 3 if there is no text before the opening `{`
```csharp
new Thing
{
    Param1 = 1,
    Param2 = 2,
}
```

So for the mappings, that leaves us with a couple options:

```
            var booleans = new Dictionary<string, Dictionary<string,bool>> 
            {
                ["True"] = new Dictionary<string, bool>
                {
                    ["true"] = true,
                }
                ,
                ["False"] = new Dictionary<string, bool>
                {
                    ["false"] = false,
                },
            };
```
```
            var booleans = new Dictionary<string, Dictionary<string,bool>> 
            {
                ["True"] = new Dictionary<string, bool> { ["true"] = true },
                ["False"] = new Dictionary<string, bool> { ["false"] = false },
            };
``` 
I don't like the second option because there may be several items in the inner-most Dictionary, and that line would get really long. Or we could do something like use one line if there are 1 or 0 items, and break it out into multiple lines if there are more. 

What do you think?

TheRealAmazonKendra: I'd vote for option 2

tests/end-to-end/vpc/App.cs
TheRealAmazonKendra: Fair. I think Hogan and I omitted it in python and java, but we can discuss as a follow up which version we should go with to ensure consistency. In Java it adds a lot of overhead so maybe we don't actually need to be consistent and just go with what's best to each language, but that still means it can be a fast follow up.

TheRealAmazonKendra:
Once Hogan and Clippy are happy, I'm good with this.

tests/end-to-end/simple/App.cs
HBobertz: option 2 seems fine to me

tests/end-to-end/vpc/App.cs
HBobertz: I think omitting it is ideal, but if we aren't consistent I think that's totally fine. Especially since I don't think we will get much traction on our non TS/Python versions

src/synthesizer/csharp/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+        for output in &ir.outputs {
+            if let Some(description) = &output.description {
+                stack_class.line("/// <summary>");
+                for description_line in description.split('\n') {
+                    stack_class.line(format!("/// {description_line}"));
+                }
+                stack_class.line("/// </summary>");
+            }
+            stack_class.line(format!("public object {} {{ get; }} ", output.name));
+            stack_class.newline();
```
iph, line 113: Note to self: check outputs to understand this :)

src/synthesizer/csharp/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+                if param.constructor_type.contains("AWS::") {
+                    let value_as = match &param.constructor_type {
+                        t if t.contains("List") => "ValueAsList",
+                        _ => "ValueAsString",
+                    };
+                    let cfn_param = ctor.indent_with_options(IndentOptions {
+                        indent: INDENT,
+                        leading: Some(
+                            format!(
+                                "props.{name} = new CfnParameter(this, \"{name}\", new CfnParameterProps\n{{")
```
iph, line 149: interesante. I thought in typescript we change parameters -> props (as that's the `cdk way`. Is that not what we are doing in all languages?

src/synthesizer/csharp/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+                        trailing: Some(format!("}}).{value_as};").into()),
+                        trailing_newline: true,
+                    });
+                    cfn_param.line(format!("Type = \"{}\",", param.constructor_type));
+                    let cfn_param_default_pre = match &param.constructor_type {
+                        t if t.contains("List") => "string.Join(\",\", ",
+                        _ => "",
+                    };
+                    let cfn_param_default_post = match &param.constructor_type {
+                        t if t.contains("List") => ")",
```
iph, line 161: Since names can be reused in Rust, we can def use shorter ones and reuse var names.

Thoughts on alternative names:
```
list_optional_prefix / list_optional_suffix
list_preamble / list_postamble
```

I'm also not the greatest wordsmith 😂 -- but when reading down below, I was like `that's verbose and I don't know what it is when reading it` which bugged me (and why I like adding the type of content in the name.

src/synthesizer/csharp/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+
+        if self.path.len() > 1 {
+            for submodule_part in self.path[1].split('-') {
+                parts.push(match submodule_part {
+                    "aws" => "AWS".into(),
+                    // TODO - This is hardcoded for now.
+                    // This part of the namespace needs to be pulled from the jsiirc.json
+                    // of the submodule. In C# there is no consistent rule we can apply to transform
+                    // this string to have the right casing.
+                    // Some are all caps, and some are Pascal case.
```
iph, line 333: Sad day :'( 

Makes sense though.

src/synthesizer/csharp/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+
+impl OutputInstruction {
+    fn emit_cfn_output(&self, output: &CodeBuffer, export: &ResourceIr, var_name: &str) {
+        let output = output.indent_with_options(IndentOptions {
+            indent: INDENT,
+            leading: Some(
+                format!(
+                    "new CfnOutput(this, \"{}\", new CfnOutputProps {{",
+                    &self.name
+                )
```
iph, line 651: I forget, does the name there == the output name in CFN?

src/synthesizer/csharp/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+                    trailing_newline: false,
+                });
+                for item in list {
+                    item.emit_csharp(&items, root_resource);
+                    items.text(",");
+                    items.newline();
+                }
+            }
+            ResourceIr::Split(sep, str) => {
+                output.text(format!("Fn.Split(\"{sep}\", "));
```
iph, line 534: interesting. This is a divergence in other langaguages (e.g. typescript actually just calls split in native, but I guess if it's on a `ref` you would have to add more complexity 🤔

iph (commented):
Minor questions, but nothing too concerning :)

src/synthesizer/csharp/mod.rs
madeline-k: This is for parameters that use one of the "special" types like AWS::SSM::etc... The value still has to be retrieved by cloudformation from ssm, so we have to generate the CfnParameter in those cases.

src/synthesizer/csharp/mod.rs
madeline-k: Yeah, I cut a corner here and used `Fn.split` in all cases. I can update this to use native c# `.Split()` when possible

src/synthesizer/csharp/mod.rs
madeline-k: outputs don't have a name field. are you thinking of the export name?

This becomes the logical id for the output. 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/outputs-section-structure.html

src/synthesizer/csharp/mod.rs
madeline-k: Thanks for this feedback, I applied your first suggestion. Let me know if I missed part of this. I'm not sure how to apply reusing var names in this scenario.

HBobertz (approved):

tests/end-to-end/vpc/App.cs
madeline-k: Okay, if we plan to omit from all the languages, then I've omitted it in c#. We can update the others in future changes.

src/synthesizer/csharp/mod.rs
madeline-k: Ok, fixed!

TheRealAmazonKendra (approved):

HBobertz (approved):

PR #284, opened 2023-09-21
fix: CSharp uses a different namespace than init template

HBobertz:
The init'd CSharp app has a namespace conflict because the namespace in the generated app file is being created with the default namespace
```
namespace Com.Acme.Test.Simple
{
    public class Cdktest01Z00Mg0HtbkCsharpMigrateStackStackProps : StackProps
    {
    }

    /// <summary>
    /// AWS CloudFormation Sample Template SQS_With_CloudWatch_Alarms: Sample template showing how to create an SQS queue with AWS CloudWatch alarms on queue depth.
    /// </summary>
    public class Cdktest01Z00Mg0HtbkCsharpMigrateStackStack : Stack
    {
```
But the namespace of the program.cs file is using the stack name
```

namespace Cdktest01Z00Mg0HtbkCsharpMigrateStack
{
    sealed class Program
    {
        public static void Main(string[] args)
        {
            var app = new App();
            new Cdktest01Z00Mg0HtbkCsharpMigrateStackStack(app, "cdktest-01z00mg0htbk-csharp-migrate-stack", new StackProps
            {
```

I figured there were 2 ways to fix this. 
a) change the generated namespace to use the stack name
b) modify the init template to have `using namespace Com.Acme.Test.Simple`
I figured option a) was cleaner so that's what I did. I changed the implementation of CSharp to no longer use the Default/new design. It wasn't actually being used for anything so I just went ahead and deleted it and swapped out the use of the default for the stack name

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -101,7 +101,7 @@ fn main() -> anyhow::Result<()> {
         #[cfg(feature = "java")]
         "java" => Box::<Java>::default(),
         #[cfg(feature = "csharp")]
-        "csharp" => Box::<CSharp>::default(),
+        "csharp" => Box::new(CSharp {}),
         unsupported => panic!("unsupported language: {}", unsupported),
     };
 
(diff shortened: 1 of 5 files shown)


madeline-k (changes requested):
It makes sense why you did it this way, but I think we should consider the whole UX here before merging this.

Please update the PR name to fix

madeline-k (approved):

PR #289, opened 2023-09-21
fix: stackname missing "stack" at the end

HBobertz:
Fixes #


TheRealAmazonKendra (changes requested):
Noting here as I did in slack that I don't think we should make this change. We should instead correctly parse the namespace and leave the class named what the user provides here.

HBobertz:
Closing until we conclude an idiomatic way of solving this issue

PR #311, opened 2023-09-27
fix(csharp): StackProps are renamed when no stack props are present

TheRealAmazonKendra:
We should only be renaming `StackProps` when there are props present. If none are, we just won't add the structure or rename the stack props.

This is also how Java is structured.

This is part 1 of the fix but it will unblock the tests in the CLI.

### DIFF
--- a/src/synthesizer/csharp/mod.rs
+++ b/src/synthesizer/csharp/mod.rs
@@ -99,13 +99,19 @@ impl Synthesizer for CSharp {
             stack_class.newline();
         }
 
+        let props_name: String = if ir.constructor.inputs.is_empty() {
+            "StackProps".to_string()
+        } else {
+            format!("{stack_name}Props").to_string()
+        };
+
         // Constructor
         let ctor = stack_class.indent_with_options(IndentOptions {
             indent: INDENT,
             leading: Some(format!(
-                "public {}(Construct scope, string id, {}Props props = null) : base(scope, id, props)\n{{",
+                "public {}(Construct scope, string id, {} props = null) : base(scope, id, props)\n{{",
                 stack_name,
-                stack_name
+                props_name
             ).into()),
             trailing: Some("}".into()),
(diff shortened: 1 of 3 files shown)


HBobertz:
Should have provided more context on my pr https://github.com/cdklabs/cdk-from-cfn/pull/279/files. Can you explain how you plan to handle the stack props in the CDK CLI?

HBobertz (changes requested):
Pulled down the PR. Creates the previous error still: 
```
/Users/bobertzh/work/deleteme/temp/myTemp/src/MyTemp/Program.cs(13,44): error CS1503: Argument 3: cannot convert from 'Amazon.CDK.StackProps' to 'MyTempStack.MyTempStackProps' [/Users/bobertzh/work/deleteme/temp/myTemp/src/MyTemp/MyTemp.csproj]
```

If we are going to implement java's prop overrides instead of using an empty props class can we do that here?

TheRealAmazonKendra:
> Pulled down the PR. Creates the previous error:
> 
> ```
> /Users/bobertzh/work/deleteme/temp/myTemp/src/MyTemp/Program.cs(13,44): error CS1503: Argument 3: cannot convert from 'Amazon.CDK.StackProps' to 'MyTempStack.MyTempStackProps' [/Users/bobertzh/work/deleteme/temp/myTemp/src/MyTemp/MyTemp.csproj]
> ```
> 
> If we are going to implement java's prop overrides instead of using an empty props class can we do that here as well?

Yes, I did the half here that would fix the use case we're testing against in the CLI integ tests to unblock upgrading on that end and then was working on doing the other half as a fast follow up.

HBobertz:
I am going to take a harder line and say this is also not consistent with CSharp documentation https://docs.aws.amazon.com/cdk/v2/guide/stack_how_to_create_multiple_stacks.html#cdk-how-to-create-multiple-stacks-extend-stackprops

PR #315, opened 2023-09-28
fix: GetAtt functions may include "." and cdk functions do not

TheRealAmazonKendra:
Line 245 in the template contains a Fn::GetAtt with 'Compliance.Type' and it is not handled correctly in any language.

The tests are being left out of this PR because not all issues have been resolved yet. The template is included for context.

Fixes #

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -111,7 +111,7 @@ dependencies = [
  "anyhow",
  "base64",
  "clap",
- "indexmap 2.0.0",
+ "indexmap 2.0.1",
  "nom",
  "numberkit",
  "phf",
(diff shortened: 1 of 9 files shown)


madeline-k (commented):
Is there any way we can remove the `.` in the IR instead? Or would that break something else?

HBobertz:
> Is there any way we can remove the `.` in the IR instead? Or would that break something else?

I'm afraid this could have some issues with our ability to separate on that . in the future if we don't already. I feel like it may be safer to do this for now imo

TheRealAmazonKendra:
I agree that we should be handing this in IR and it also makes me realize that there's a bunch of stuff we're handing in individual languages that should probably be done in IR instead. Anyway, fixed it here.

madeline-k (approved):

Issue #319, opened 2023-09-28
python: lambda imports fail


TheRealAmazonKendra:
lambda seems to be a protected term in python because `import aws_cdk.lambda as lambda` cannot find the import. The error resolved when this is changed to `import aws_cdk.aws_lambda as aws_lambda`. We may want to consider prefixing all our imports with `aws_` in case there are other cases of this we have not yet discovered.

PR #323, opened 2023-09-28
fix(python): python reserved keywords overlapping with service names

HBobertz:
Previously python apps would fail lambda because we would be using the keyword lambda. i.e.
```
import aws_cdk.aws_lambda as lambda

lambdaFunction = lambda.CfnFunction(self, 'LambdaFunction'
```

fixed by only adding `aws_` before a lambda import
```
import aws_cdk.aws_lambda as aws_lambda

lambdaFunction = aws_lambda.CfnFunction(self, 'LambdaFunction'
```

Fixes # https://github.com/cdklabs/cdk-from-cfn/issues/319

### DIFF
--- a/src/synthesizer/python/mod.rs
+++ b/src/synthesizer/python/mod.rs
@@ -244,10 +244,9 @@ impl ImportInstruction {
         let module = parts.join(".");
         if !module.is_empty() {
             // "lambda" is a reserved keyword in python, so we need to change it to aws_lambda
-            if self.name == "lambda"{
+            if self.name == "lambda" {
                 format!("import {} as aws_{}", module, self.name,)
-            }
-            else {
+            } else {
                 format!("import {} as {}", module, self.name,)
             }
         } else {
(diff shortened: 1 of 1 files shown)


TheRealAmazonKendra (changes requested):
I think we should use the `aws_` convention across the board for consistency. There may be other services we haven't discovered yet that are also reserved keywords.

TheRealAmazonKendra:
Also, please link the issue you're fixing in the PR body.

HBobertz:
> I think we should use the `aws_` convention across the board for consistency. There may be other services we haven't discovered yet that are also reserved keywords.

https://docs.aws.amazon.com/cdk/v2/guide/hello_world.html#hello_world_tutorial_add_bucket

I think this is more aligned with how we do this in python. So I'd personally like to keep it

TheRealAmazonKendra:
Do we know for sure that other service names aren't reserved keywords? Making the change on the whole safeguards us from that being an issue for other services. When that example was written, that issue probably wasn't in the forefront of their mind

HBobertz:
> Do we know for sure that other service names aren't reserved keywords? Making the change on the whole safeguards us from that being an issue for other services. When that example was written, that issue probably wasn't in the forefront of their mind.

https://docs.python.org/2.5/ref/keywords.html 
https://docs.aws.amazon.com/

Compared the list of Python keywords and the AWS services. There doesn't seem to be any other keyword overlap between the two.

madeline-k (approved):

Issue #318, opened 2023-09-28
all languages: type not properly translated into booleans


TheRealAmazonKendra:
When a template has the following:
```
"Type": "String",
 "Default": "false",
 "AllowedValues": ["false", "true"]
```
The CDK translates this into a boolean but cdk-from-cfn translates it into a string across all languages. This is causing every language to fail for the parameters block in [this template](https://github.com/cdklabs/cdk-from-cfn/pull/315/files#diff-16af47ca157647af1f02c784ec1666ee56accbcbbe84caccb8f1aa6533957f45)

PR #326, opened 2023-09-29
fix(all): booleans are processed as strings causing failures

TheRealAmazonKendra:
When "AllowedValues" in the CloudFormation template is populated and the allowed values are "true" and "false", the CDK translates these props into a boolean instead of a string. However, when other allowed values are present, the prop should still be represented as a string. 

As before, I have not included the tests, as before, because they're not fully passing yet but I've attached the template that I'm working against for reference. This template contains an example of allowed values that should resolve to a string, while the last template I uploaded translates allowed values that should resolve to a boolean.

Fixes https://github.com/cdklabs/cdk-from-cfn/issues/318

### DIFF
--- a/src/ir/constructor/mod.rs
+++ b/src/ir/constructor/mod.rs
@@ -14,8 +14,27 @@ impl Constructor {
                 .map(|(name, param)| ConstructorParameter {
                     name: camel_case(&name),
                     description: param.description,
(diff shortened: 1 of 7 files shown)


madeline-k (approved):

madeline-k:
I like the branch name for this one

PR #329, opened 2023-09-29
fix(python): python outputs weren't very pythonic

HBobertz:
Currently python outputs were being both emitted as a class variable at the bottom of the class and as a global variable at the top of the class. I chose to emit them as globals to make it obvious what their purpose was and to pretend python has some consistency with other languages which require you declare your class variables before assigning their values in the constructor. This was not a pythonic way of doing this so I'm removing it and instead I'm just annotating the pydoc descriptions to the class variables themselves. 

### DIFF
--- a/src/synthesizer/python/mod.rs
+++ b/src/synthesizer/python/mod.rs
@@ -168,7 +168,6 @@ impl Synthesizer for Python {
                     let comment = ctor.pydoc();
                     comment.line(description.to_owned());
                 }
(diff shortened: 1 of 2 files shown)


TheRealAmazonKendra:
Reminder: a fix should describe the problem, not the fix. Also, we should include the language in the title like `fix(python): blah blah blah`

TheRealAmazonKendra (approved):

PR #333, opened 2023-09-29
fix(csharp): bool cannot use ??=

TheRealAmazonKendra:
The bool must be nullable. Thanks for the assist @madeline-k!

Now that the tests will pass and the code synths, I've added the C# ones in this PR.

Fixes #


madeline-k (approved):
🎉

PR #336, opened 2023-09-29
fix(python): camel case where there should be snake case

HBobertz:
Fixes improper naming scheme for properties

### DIFF
--- a/src/synthesizer/python/mod.rs
+++ b/src/synthesizer/python/mod.rs
@@ -416,9 +416,9 @@ impl Reference {
             Origin::PseudoParameter(x) => match x {
                 PseudoParameter::Partition => "self.partition".into(),
                 PseudoParameter::Region => "self.region".into(),
-                PseudoParameter::StackId => "self.stackId".into(),
-                PseudoParameter::StackName => "self.stackName".into(),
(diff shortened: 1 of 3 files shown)


src/synthesizer/python/mod.rs
```diff
@@ -416,9 +416,9 @@ impl Reference {
             Origin::PseudoParameter(x) => match x {
                 PseudoParameter::Partition => "self.partition".into(),
                 PseudoParameter::Region => "self.region".into(),
-                PseudoParameter::StackId => "self.stackId".into(),
-                PseudoParameter::StackName => "self.stackName".into(),
-                PseudoParameter::URLSuffix => "self.urlSuffix".into(),
+                PseudoParameter::StackId => "self.stack_id".into(),
+                PseudoParameter::StackName => "self.stack_name".into(),
+                PseudoParameter::URLSuffix => "self.url_suffix".into(),
```
TheRealAmazonKendra, line 421: I know this is still in draft but just adding this question so I don't forget when I review next week. Does `self.notificationArns` also need a casing change?

HBobertz: yup good catch

TheRealAmazonKendra (approved):

PR #340, opened 2023-09-30
fix(python): Python props were ugly and not very pythonic

HBobertz:
I realized I didn't need a ternary operator to default these values and that the second positional argument of kwargs.get is a built in default value

### DIFF
--- a/src/synthesizer/python/mod.rs
+++ b/src/synthesizer/python/mod.rs
@@ -78,6 +78,8 @@ impl Synthesizer for Python {
             .collect::<Vec<&ConstructorParameter>>();
         if !have_default_or_special_type_params.is_empty() {
             ctor.newline();
+            // props are handled weirdly in python. Python doesn't have interfaces so we try and retrieve
+            // the props from kwargs, and default to None or a default value if one is given.
             ctor.line("# Applying default props");
             let obj = ctor.indent_with_options(IndentOptions {
                 indent: INDENT,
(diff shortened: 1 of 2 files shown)


src/synthesizer/python/mod.rs
```diff
@@ -18,6 +18,8 @@ use super::Synthesizer;
 
 const INDENT: Cow<'static, str> = Cow::Borrowed("  ");
 
+// reserved python keywords as of 3.13 (https://docs.python.org/3/reference/lexical_analysis.html#keywords)
+// if we happen to name a module with one of these keywords, we need to prepend 'aws_' to avoid a name conflict
```
HBobertz, line 22: decided to just throw in this comment while I was at it. LMK if you want me to put this in it's own pr

src/synthesizer/python/mod.rs
```diff
@@ -85,7 +89,6 @@ impl Synthesizer for Python {
             });
             for param in have_default_or_special_type_params {
                 let name = &param.name;
-                // example: AWS::EC2::Image::Id, List<AWS::EC2::VPC::Id>, AWS::SSM::Parameter::Value<List<String>>
```
HBobertz, line 88: this honestly isn't very helpful, I just decided to remove it

TheRealAmazonKendra: I've actually found this comment helpful. Can we please leave it in?

HBobertz: oh sure, yeah re-added it

TheRealAmazonKendra (approved):

PR #345, opened 2023-10-02
fix(typescript): exported type mismatch

HBobertz:
Fixes: exported value only expects a string but we would give it any type. Fixing this by casting it to a string

```
Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'
```

### DIFF
--- a/tests/end-to-end/simple/app.ts
+++ b/tests/end-to-end/simple/app.ts
@@ -147,7 +147,7 @@ export class SimpleStack extends cdk.Stack {
       new cdk.CfnOutput(this, 'BucketArn', {
         description: 'The ARN of the bucket in this template!',
         exportName: 'ExportName',
-        value: this.bucketArn,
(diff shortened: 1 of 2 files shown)


tests/end-to-end/simple/app.ts
```diff
@@ -147,7 +147,7 @@ export class SimpleStack extends cdk.Stack {
       new cdk.CfnOutput(this, 'BucketArn', {
         description: 'The ARN of the bucket in this template!',
         exportName: 'ExportName',
-        value: this.bucketArn,
+        value: this.bucketArn?.toString() ?? '',
```
TheRealAmazonKendra, line 150: We should assert that this will definitely be there, not to allow it to be blank

TheRealAmazonKendra (changes requested):

TheRealAmazonKendra (approved):

PR #352, opened 2023-10-04
fix(python): pretty_name is not snake_case

TheRealAmazonKendra:
The function pretty_name is not handling the casing correctly. This change uses the snake_case function to correct this.

This also includes the test cases I've been checking against with all the python changes. With this last change, they now synth!

After my three changes get merged, I will add the code that actually runs these new tests.

Fixes #


TheRealAmazonKendra:
Linter is failing because of extra import but I'm leaving it so it doesn't cause a conflict when #351 is merged.

HBobertz (dismissed):
love the branch name

HBobertz (approved):

PR #389, opened 2023-10-27
Fix a bug parsing files with ! operator

Pr0methean:
I have a Cfn template that uses `!` as a type specifier before strings that contain newlines. For example:
```
            /etc/cfn/cfn-hup.conf:
              content:
                Fn::Join:
                - ''
                - - ! '[main]

'
                  - stack=
                  - Ref: AWS::StackId
                  - ! '

'
                  - region=
                  - Ref: AWS::Region
                  - ! '

'
              group: root
              mode: '000444'
              owner: root
```

That leads to errors like these:
```
Error: Resources.ArtifactEncryptionKey.Properties.KeyPolicy.Statement[0].Principal.AWS.Fn::Join[1][0]: unknown variant `!`, expected one of `Base64`, `Cidr`, `FindInMap`, `GetAtt`, `GetAZs`, `If`, `ImportValue`, `Join`, `Select`, `Split`, `Sub`, `Ref` at line 205 column 19
```
This PR fixes those errors by changing the return type of `IntrinsicFunction::from_enum` so that it can wrap the next token instead of an IntrinsicFunction, and adding a match case for "!" that does so.

### DIFF
--- a/src/parser/intrinsics/mod.rs
+++ b/src/parser/intrinsics/mod.rs
@@ -68,7 +68,7 @@ static INTRINSIC_FUNCTION_TAGS: &[&str] = &[
 ];
 
 impl IntrinsicFunction {
-    pub(super) fn from_enum<'de, A: serde::de::EnumAccess<'de>>(data: A) -> Result<Self, A::Error> {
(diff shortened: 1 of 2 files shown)


TheRealAmazonKendra (changes requested):
Thank you for your contribution. Please add a test case that covers this change and make sure the code is passing the linter.

PR #427, opened 2023-11-29
feat: extend end-to-end test workflows

madeline-k:
This PR extends the current end-to-end tests to add some additional capabilities. See the [workflow diagram](https://github.com/cdklabs/cdk-from-cfn/blob/8f702cebfdf6a9b75d965bbeaeb96ec63122f975/tests/end-to-end/README.md#workflow-design-what-happens-when-you-run-the-tests) for the full flow. The high-level steps for each end-to-end test are:
1. Create a Cloudformation Stack from the input CFN template
2. Run cdk-from-cfn to synthesize a cdk stack definition in any target language from the input CFN template
3. Create a working CDK App with the cdk stack definition
4. Run CDK synth on the CDK App
5. Verify the output of CDK synth has not changed, and store the diff from the original template in a file for review.

### Notes for reviewers:
- I moved some logic out of the test_case macro and into functions to make it easier to work with, because auto-complete does not work inside macros. The macro is still necessary to avoid having to write a test module for each end-to-end test and a test fn for each language. https://github.com/rust-lang/rust-analyzer/issues/12524
- Using environment variables instead of cli flags to configure the end to end test workflow, because passing cli args through to your test functions is not supported by Rust's libtest. (We could do this with a custom test binary implementation as a future improvement.)
- I am keeping the generated CDK stack code in the "snapshots", because this has been very helpful for reviewing changes to the code-gen. With this update, manual review of the generated CDK code is no longer always necessary to verify that the generated code works, because the end-to-end tests now actually run `cdk synth` on them. The manual review is especially helpful for reviewing style, format, etc. of the generated code.
- CDK Synth is being skipped for some tests for a few reasons
  - simple::golang
    <details>
    <summary>produces code that has errors during cdk synth (expand to see errors)</summary>
    
    ```
    ./stack.go:100:2: isUs declared and not used
    ./stack.go:100:71: not enough arguments in call to cdk.Fn_Split
        have (*string, *string)
        want (*string, *string, *float64)
    ./stack.go:129:27: undefined: LoggingConfiguration
    ./stack.go:132:27: undefined: WebsiteConfiguration
    ./stack.go:133:29: undefined: RedirectAllRequestsTo
    ./stack.go:143:21: cannot use table[jsii.String("Values")][jsii.String("String")] (map index expression of type interface{}) as *string value in argument to cdk.Fn_Base64: need type assertion
    ```
    
    </details>
  
  - config::golang
      <details>
    <summary>produces code that has errors during cdk synth (expand to see errors)</summary>
        
        ```
        ./stack.go:86:18: undefined: Policy
        ./stack.go:87:6: undefined: Policy
        ./stack.go:131:18: undefined: Policy
        ./stack.go:132:6: undefined: Policy
        ./stack.go:201:39: undefined: ConfigSnapshotDeliveryProperties
        ./stack.go:213:11: undefined: Code
        ./stack.go:271:21: undefined: RecordingGroup
        ./stack.go:285:12: undefined: Scope
        ./stack.go:291:13: undefined: Source
        ./stack.go:293:24: undefined: SourceDetail
        ./stack.go:293:24: too many errors
        ```
        </details>
  - documentdb::golang
      <details>
    <summary>produces code that has errors during cdk synth (expand to see errors)</summary>
       
      ```
      ./stack.go:5:2: "github.com/aws/aws-cdk-go/awscdk/v2/awsdocdb" imported as docdb and not used
      ./stack.go:40:15: undefined: doc_db
      ./stack.go:43:4: undefined: doc_db
      ./stack.go:51:2: undefined: doc_db
      ./stack.go:54:4: undefined: doc_db
      ```
      
      </details>
  
  - resource_w_json_type_properties::golang
      <details>
    <summary>produces code that has errors during cdk synth (expand to see errors)</summary>
      
      ```
      ./stack.go:44:18: undefined: Policy
      ./stack.go:45:6: undefined: Policy
      ```
      
      </details>
  
  - config::java, resource_w_json_type_properties::java  
    These tests do not fail CDK synth, but CDK synth produces templates with a non-deterministic order. This makes the tests validation of the produced template fail. We need to make the validation mechanism for these tests more sophisticated so that it can recognize equivalent templates, even when the json objects and properties are not necessarily in the same order.
  - sam_nodejs_lambda and sam_nodejs_lambda_arr_transform
     These are new tests since I've been finalizing this PR. They still need to be run with `cdk synth` to see if it works or not.

### Future improvements to the end-to-end tests
https://github.com/cdklabs/cdk-from-cfn/issues/506
https://github.com/cdklabs/cdk-from-cfn/issues/507
https://github.com/cdklabs/cdk-from-cfn/issues/508
https://github.com/cdklabs/cdk-from-cfn/issues/509
https://github.com/cdklabs/cdk-from-cfn/issues/510 
https://github.com/cdklabs/cdk-from-cfn/issues/511


tests/end-to-end.rs
```diff
@@ -22,86 +41,366 @@ macro_rules! test_case {
...
+            trailing: Some("}),".into()),
+            trailing_newline: true,
+        });
+        app_props.line("generateBootstrapVersionRule: false,");
+        code.line(format!("new {}(app, \"Stack\");", cdk_stack_classname));
+        code.line("app.synth();");
+        code.write(&mut file).unwrap();
+    } else {
+        // Fail the test to prevent tests without snapshots from succeeding in CI/CD
+        panic!("CDK App file not found at {}. If you are developing a new test, set UPDATE_SNAPSHOTS=true in your environment variables and the test will create a default app file.", cdk_app_filename);
```
comcalvi, line 292: This will make the test fail with a diff between a default app and whatever was synthesized. Would it be preferable to fail with a diff between an empty file and the code that was synthesized?

tests/end-to-end.rs
```diff
@@ -22,86 +41,366 @@ macro_rules! test_case {
...
+                &language_boilerplate_dir,
+                &test_working_dir,
+                &expected_outputs_dir,
+                &mut snapshots_zip,
+            );
+
+            diff_original_template_with_new_templates(stringify!($name), &test_working_dir);
+
+            // update snapshots
+            update_snapshots(
```
comcalvi, line 121: options type for this? If it's only ever called in one place this is fine, but if we have to call it more than once the options type would make this easier

tests/end-to-end.rs
```diff
@@ -22,86 +41,366 @@ macro_rules! test_case {
...
+    )
+    .unwrap();
+
+    // Template and diff files
+    let walkdir = WalkDir::new(test_working_dir);
+    for entry in walkdir.into_iter().map(|e| e.unwrap()) {
+        let filename = entry.file_name().to_str().unwrap();
+        if entry.path().to_str().unwrap().contains("node_modules") {
+            continue;
+        }
```
comcalvi, line 390: Our snaps can be non-JS/TS, right? We should also exclude any folders created by other languages for dependencies

tests/end-to-end/simple/typescript/Stack.diff
```diff
@@ -0,0 +1,276 @@
+diff --git a/./tests/end-to-end/simple/template.json b/tests/end-to-end/simple-typescript-working-dir/cdk.out/Stack.template.json
```
comcalvi, line 1: leftover?

comcalvi (commented):
Looks really good! I have mostly minor comments

tests/end-to-end.rs
```diff
@@ -22,86 +41,366 @@ macro_rules! test_case {
...
+    let id = resp.stack_id.unwrap_or_default();
+    print!("Stack {id} create in progress...");
+    std::io::stdout().flush().unwrap();
+
+    let mut status = check_stack_status(&id, &client).await.unwrap();
+
+    while let StackStatus::CreateInProgress = status {
+        print!(".");
+        std::io::stdout().flush().unwrap();
+        tokio::time::sleep(std::time::Duration::new(2, 0)).await;
```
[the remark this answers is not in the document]
madeline-k: This just prints a new dot every 2 seconds. Most stacks will not be done in 2 seconds. I can increase this interval if you'd like though.

tests/end-to-end.rs
```diff
@@ -22,86 +41,366 @@ macro_rules! test_case {
...
+        .create_stack()
+        .stack_name(stack_name)
+        .template_body(template)
+        .on_failure(OnFailure::Delete)
+        .send()
+        .await
+        .unwrap();
+    let id = resp.stack_id.unwrap_or_default();
+    print!("Stack {id} create in progress...");
+    std::io::stdout().flush().unwrap();
```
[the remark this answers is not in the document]
madeline-k: the unwrap() will make the test crash if the flush fails. Without unwrap, we would have to manually handle the result and check if there was an error.

tests/end-to-end.rs
```diff
@@ -22,86 +41,366 @@ macro_rules! test_case {
...
+            let test_working_dir = format!(
+                "tests/end-to-end/{}-{}-working-dir",
+                stringify!($name),
+                stringify!($lang)
+            );
+            let language_boilerplate_dir = format!(
+                "tests/end-to-end/app-boilerplate-files/{}",
+                stringify!($lang)
+            );
+            synth_cdk_app(
```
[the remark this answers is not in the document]
madeline-k: I am considering a couple refactors for this with @MrArnoldPalmer as well. Will take this one into consideration.

tests/end-to-end.rs
```diff
@@ -22,86 +41,366 @@ macro_rules! test_case {
...
+    }
+}
+
+async fn check_stack_status(id: &str, client: &Client) -> Result<StackStatus, Error> {
+    println!("Checking stack status: {}", id);
+    let resp = client.describe_stacks().stack_name(id).send().await?;
+    if let Some(stacks) = resp.stacks {
+        if let Some(stack) = stacks.first() {
+            if let Some(status) = &stack.stack_status {
+                return Ok(status.clone());
```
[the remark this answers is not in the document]
madeline-k: This function owns the memory for `status`. In order to return it from the function, I need to make a copy. I am still pretty new to Rust, so let me know if there is a better approach for that!

tests/end-to-end.rs
```diff
@@ -22,86 +41,366 @@ macro_rules! test_case {
...
+    let app_dst_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join(test_working_dir)
+        .join(cdk_app_filename);
+    println!("DEBUG: app_dst_path: {:?}", app_dst_path);
+    if let Ok(mut cdk_app_file) = snapshots_zip.by_name(&cdk_app_file_path) {
+        let mut contents = Vec::<u8>::new();
+        cdk_app_file.read_to_end(&mut contents).unwrap();
+        println!("An app file already exists. Copying it to: {:?}", app_dst_path);
+        let mut file = File::create(app_dst_path).unwrap();
+        file.write_all(contents.as_bytes()).unwrap();
```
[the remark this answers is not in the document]
madeline-k: yes exactly. We don't want to proceed with the test if any of these file system operations fail.

tests/end-to-end.rs
```diff
@@ -22,86 +41,366 @@ macro_rules! test_case {
...
+    } else if std::env::var_os("UPDATE_SNAPSHOTS").is_some() {
+        // If it does not already exist, create new app file
+        println!("UPDATE_SNAPSHOTS=true, and there is no existing app file, creating default one at {}...", app_dst_path.to_str().unwrap());
+        let mut file = File::create(app_dst_path).unwrap();
+        let code: CodeBuffer = CodeBuffer::default();
+        // Recommend to manually instantiate a stack for each combination of parameters/conditionals that should be tested
+        code.line("// auto-generated! a human should update this!");
+        code.line("import * as cdk from \"aws-cdk-lib\";");
+        code.line(format!(
+            "import {{ {} }} from \"./stack\";",
```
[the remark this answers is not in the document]
madeline-k: It's inside a `format!` macro. To actually render a `{` you have to escape it like this: `{{`. the single `{}` is for inserting a value into the string.

tests/end-to-end.rs
```diff
@@ -22,86 +41,366 @@ macro_rules! test_case {
...
+            trailing: Some("});".into()),
+            trailing_newline: true,
+        });
+        let app_props = app.indent_with_options(IndentOptions {
+            indent: INDENT,
+            leading: Some("defaultStackSynthesizer: new cdk.DefaultStackSynthesizer({".into()),
+            trailing: Some("}),".into()),
+            trailing_newline: true,
+        });
+        app_props.line("generateBootstrapVersionRule: false,");
```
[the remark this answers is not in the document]
madeline-k: to minimize the boilerplate stuff that will add to the diff between the original template, and the template generated by the CDK stack generated by cdk-from-cfn.

tests/end-to-end.rs
madeline-k: The `App.ts` file is not part of the snapshot diffing stuff. It is just there so that we have a complete app to synth. And so we can test all combinations of parameters to the stack. 

I don't really understand your question. Can you clarify?

tests/end-to-end.rs
madeline-k: yes, that is coming very soon!

tests/end-to-end.rs
comcalvi: ah I thought this was part of the snapshot diff, as in the base expected template. I'm not really sure why we need this to have a complete app to synth, could you elaborate a bit on why we need empty CDK apps on disk?

tests/end-to-end.rs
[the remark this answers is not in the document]
comcalvi: ahh I see. No need to increase this interval, this is fine

tests/end-to-end.rs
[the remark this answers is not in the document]
comcalvi: I don't really like this pattern in rust, but I think this is how the language wants us to do this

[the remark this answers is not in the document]
MrArnoldPalmer: I'd recommend using `.expect` over `.unwrap` wherever possible to make panics on `Err`s more explicit. It allows you to add some custom messaging as well.

src/synthesizer/golang/mod.rs
madeline-k, line 1: The changes in this file are replicated in https://github.com/cdklabs/cdk-from-cfn/pull/452 When that PR is merged, these will be removed from this one.

tests/end-to-end/README.md
```diff
@@ -0,0 +1,25 @@
+How to add a new test
+
+
+1. Create a folder under end-to-end with the name of your new test
+
+
+fail with error and tell developer to turn on
```
comcalvi, line 7: missing number 2 here?

tests/end-to-end.rs
```diff
@@ -1,107 +1,735 @@
...
+        if self.skip_cdk_synth {
+            self.update_cdk_stack_def_snapshot(&cdk_stack_definition);
+        } else {
+            self.synth_cdk_app(&cdk_stack_definition);
+
+            self.diff_original_template_with_new_templates();
+
+            self.check_new_templates_and_diffs_match_expected();
+
+            self.update_snapshots(&cdk_stack_definition);
```
comcalvi, line 176: check_new_templates... and update_snapshots both exit immediately depending on the value `UPDATE_SNAPSHOTS`. Do we need the early return within those functions or can we pull those checks and never call them to begin with? It would make the control flow a bit easier to follow.

build.rs
```diff
@@ -206,3 +214,50 @@ impl PropertyRule {
...
+        let path = entry.path();
+        let name = path
+            .strip_prefix(Path::new(src_dir))
+            .expect(&format!("{src_dir} should be a prefix of {:?}", path))
+            .to_str()
+            .expect("failed to convert filename to string");
+
+        for d in do_not_zip {
+            if name.contains(d) {
+                continue 'dir_entries;
```
comcalvi, line 242: do we need this label?

comcalvi (commented):
nothing but minor comments, this looks really solid!

build.rs
madeline-k: yes, because at this point I want to continue on the outer loop, not this inner one.

tests/end-to-end.rs
madeline-k: yes, I will make that change if it's easier to read that way!

tests/end-to-end/README.md
madeline-k: ah yes I will update this, thanks!

comcalvi (dismissed):

tests/end-to-end/README.md
```diff
@@ -0,0 +1,582 @@
...
+      CloudFormation template
+        - `tests/end-to-end/mytest/<language>/Stack.diff`
+            - The name of this file will match the template file, and may vary
+              if you edited the app file to give the stack a different name.
+            - If this file does not exist, that means the new stack is identical
+              to the original stack. Hooray! You do not need to manually review
+              to see if the changes are acceptable.
+    
+    Note: This diff mechanism should be improved so that this manual review is
+    either not necessary, or to make it easier. See [issue](TODO add link).
```
TheRealAmazonKendra, line 239: Missing link here

TheRealAmazonKendra (changes requested):

issues.md
```diff
@@ -0,0 +1,8 @@
+End-to-end tests: Use a more advanced mechanism to compare CloudFormation templates
```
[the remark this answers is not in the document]
madeline-k: oh thanks for catching, this was just personal notes for me. I've created all the actual issues, and linked them from the PR description.

TheRealAmazonKendra (approved):

PR #450, opened 2023-12-18
fix(csharp): if user does not supply props to stack constructor, they remain null and cause a C# runtime error

madeline-k:
In this case, set props to default constructor.

### DIFF
--- a/src/synthesizer/csharp/mod.rs
+++ b/src/synthesizer/csharp/mod.rs
@@ -118,6 +118,7 @@ impl Synthesizer for CSharp {
             .collect::<Vec<&ConstructorParameter>>();
         if !have_default_or_special_type_params.is_empty() {
             ctor.line("// Applying default props");
+            ctor.line("props ??= new SimpleStackProps();");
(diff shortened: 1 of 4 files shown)


TheRealAmazonKendra:
For these cases, I think that this is expected. These would both require a user to enter values.

HBobertz (approved):

HBobertz:
Hm actually I thought it was fine then realized maybe it's not. Is this not finding any props? Or does this just fail when a single prop can't be found?

madeline-k:
This change enables you to write app code like [this](https://github.com/cdklabs/cdk-from-cfn/pull/427/files?short_path=fb70f22#diff-4e34ae7121d3e3118c174c4738607f3cd80c34d51ea074d71c5dd58c7543bbccR14)

```
new SimpleStack.SimpleStack(app, "Stack");
```

Where the props object is completely omitted. In this example, the props are optional, and all props have default values. But without this change, it would still cause a runtime error. Because the default StackProps that need to be passed to the base Stack class constructor are not instantiated.

PR #452, opened 2023-12-20
fix(go): cdk-from-cfn does not produce working go files

madeline-k:
This fixes 3 issues in order to allow the most basic of go stacks to work. It does not fix all current issues with go, so the title might be a bit too broad. If anyone has suggestions for a better title here, I'd love to here it.

| | Issue  | Fix |
|-| ------------- | ------------- |
|1| A file produced by cdk-from-cfn cannot be used as the main executable in a CDK app because it declares a `main` function in a package that is not called main.  | change the package name line to `package main`. See https://go.dev/doc/modules/layout |
|2| The props parameter of the Stack constructor is not a pointer, which prevents it from being able to be optional. | Make props a pointer.  |
|3| If `nil` is supplied as the prop argument of the Stack constructor, a runtime error occurs because we cannot dereference `nil` to get the `StackProps` to pass to the Stack base class constructor    | Handle nil case, and create default `StackProps`                        |

### DIFF
--- a/src/synthesizer/golang/mod.rs
+++ b/src/synthesizer/golang/mod.rs
@@ -47,7 +47,7 @@ impl Synthesizer for Golang {
     ) -> io::Result<()> {
         let code = CodeBuffer::default();
 
-        code.line(format!("package {}", self.package_name));
+        code.line("package main");
         code.newline();
 
         let imports = code.indent_with_options(IndentOptions {
@@ -107,7 +107,7 @@ impl Synthesizer for Golang {
             indent: INDENT,
             leading: Some(
                 format!(
-                    "func New{}(scope constructs.Construct, id string, props {}Props) *{} {{",
+                    "func New{}(scope constructs.Construct, id string, props *{}Props) *{} {{",
                     stack_name, stack_name, stack_name
                 )
                 .into(),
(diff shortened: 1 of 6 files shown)


src/synthesizer/golang/mod.rs
```diff
@@ -47,7 +47,7 @@ impl Synthesizer for Golang {
     ) -> io::Result<()> {
         let code = CodeBuffer::default();
 
-        code.line(format!("package {}", self.package_name));
+        code.line("package main");
```
madeline-k, line 50: This is fix #1 from description

src/synthesizer/golang/mod.rs
```diff
@@ -107,7 +107,7 @@ impl Synthesizer for Golang {
             indent: INDENT,
             leading: Some(
                 format!(
-                    "func New{}(scope constructs.Construct, id string, props {}Props) *{} {{",
+                    "func New{}(scope constructs.Construct, id string, props *{}Props) *{} {{",
```
madeline-k, line 110: This is fix # 2 from description

src/synthesizer/golang/mod.rs
```diff
@@ -196,8 +196,15 @@ impl Synthesizer for Golang {
...
-        ctor.line("stack := cdk.NewStack(scope, &id, &props.StackProps)");
+        ctor.line("var sprops cdk.StackProps");
+        let props_not_nil_block = ctor.indent_with_options(IndentOptions {
+            indent: INDENT,
+            leading: Some("if props != nil {".into()),
+            trailing: Some("}".into()),
+            trailing_newline: true,
+        });
+        props_not_nil_block.line("sprops = props.StackProps");
+        ctor.line("stack := cdk.NewStack(scope, &id, &sprops)");
```
madeline-k, line 207: This is fix # 3 from description

TheRealAmazonKendra (approved):

PR #529, opened 2024-01-31
feat: use cdk patched schema to determine resource and property types

TheRealAmazonKendra:
There are still a few TODO's here (for a future PR, this one is already too big):
1. There is ample room to simplify each of the synthesizers now. We should do this as a fast follow up.
2. I ran tests locally for the `Subnet` resource, which first identified that our mismatched types had a problem, but the extra testing files felt like they added too much extra noise to this PR. 
3. Automation of the ingestion of the parsed schema: the code for paring is written but for the sake I speed, I manually added it in the meantime.
4. Addition of union types for types that have previous versions. We track current and previous types and, in our main library, allow any of them as inputs. Here, I've defaulted to the previous type because these are usually json to something more strongly typed, which caused the bug in `Subnet`

Unexpected Benefits of this change:
1. Go is much closer to working and does synth in some cases.

Fixes #


src/cdk/schema.rs
```diff
@@ -0,0 +1,580 @@
...
+}
+
+impl<'de, V: serde::Deserialize<'de> + 'static> serde::Deserialize<'de> for Map<V> {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        struct MapVisitor<V>(PhantomData<V>);
+        impl<'de, V: serde::Deserialize<'de> + 'static> serde::de::Visitor<'de> for MapVisitor<V> {
+            type Value = Map<V>;
+
+            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+                formatter.write_str("a mapping of string to value")
```
HBobertz, line 137: this seems a bit generic

src/cdk/schema.rs
```diff
@@ -0,0 +1,580 @@
...
+    // The name of the type in Go.
+    #[cfg(feature = "java")]
+    pub java: JavaName,
+
+    // The name of the type in Python.
+    #[cfg(feature = "python")]
+    pub python: PythonName,
+}
+
+impl Clone for TypeName {
```
HBobertz, line 284: I'm 99% sure since these are all just unconditional clones you can instead just put a `#[derive(Clone)]` on the 
`pub struct TypeName {` above.

src/cdk/schema.rs
```diff
@@ -0,0 +1,580 @@
...
+}
+
+impl PropertyBag for DataType {
+    fn property(&self, name: &str) -> Option<Property> {
+        self.properties.get(name).map(Clone::clone)
+    }
+}
+
+// A multi-language aware type name.
+#[derive(serde::Deserialize)]
```
HBobertz, line 261: Think you can just put `#[derive(Clone)]` here and not do the thing below this

src/ir/resources/mod.rs
```diff
@@ -473,6 +455,30 @@ impl ResourceType {
...
+                };
+                let type_name = match parts.next() {
+                    Some("") | None => {
+                        return Err(TransmuteError::new(format!(
+                            "invalid resource type: {from:?} (missing resource type name)"
+                        )))
+                    }
+                    Some(type_name) => type_name.into(),
+                };
+                if parts.next().is_some() {
```
HBobertz, line 475: Will this fail if someone passes a hook or module? I'm not sure exactly how thes einteract but I think they have 4 segments and the last is like ::hook

HBobertz (approved):
Nothing I consider blocking, just some potentially un-idionmatic stuff that doesn't matter

HBobertz (approved):

PR #578, opened 2024-02-23
fix(csharp): cdk stacks with nested `ResourceIr` objects fail to synthesize due to implicit type conversion errors

colifran:
This PR fixes a csharp synthesizer bug that occurs when a `ResourceIr` object is nested inside a resource constructor. When this occurs, any property values associated with the nested `ResourceIr` object are incorrectly wrapped in double quotes. An example can be seen in the code snippet below. The variable `Order` was being assigned the string "1" where it should have been the integer 1. As a result, the synthesized CDK app throws the following error:

```
Cannot implicitly convert type 'string' to 'double'
```

Also, note that the variable `Priority` was correctly assigned because it is not contained within the nested `ResourceIr` object, i.e., `CfnJobQueue.ComputeEnvironmentOrderProperty`.

```csharp
var jobQueue = new CfnJobQueue(this, "JobQueue", new CfnJobQueueProps
{
    Priority = 1,
    ComputeEnvironmentOrder = new []
    {
        new CfnJobQueue.ComputeEnvironmentOrderProperty
        {
            Order = "1",
            ComputeEnvironment = computeEnvironment.Ref,
        },
    },
});
```

This bug happens while recursively calling the method `emit_csharp` which was being wrapped in double quotes for integers, doubles, and bools.

```csharp
match val {
    ResourceIr::Bool(_) | ResourceIr::Number(_) | ResourceIr::Double(_) => {
        object_block.text("\"");
        val.emit_csharp(&object_block, schema);
        object_block.text("\"");
    }
    _ => val.emit_csharp(&object_block, schema),
}
```

The fix for this is to just call `emit_csharp` for all value types without wrapping them in double quotes:

```csharp
val.emit_csharp(&object_block, schema);
```

To test this, a new end-to-end test for an AWS Batch [template](https://s3.amazonaws.com/cloudformation-templates-us-east-1/Managed_EC2_Batch_Environment.template) was added.


TheRealAmazonKendra (approved):
Looks great!

PR #585, opened 2024-02-28
fix(all): synthesizers are converting casing on ECS task definition log configuration options causing deployment failures

colifran:
This PR fixes a bug where all synthesizers are converting casing for ECS task definition log configuration options to the casing used by their language, i.e., python converts to snake case, typescript converts to camel case, etc. This is causing deployment failures because ECS task definition log configuration options must maintain a hyphenated format. For example, the keys in the options object below must stay hyphenated.

```ts
logConfiguration: {
  options: {
    'awslogs-group': '/aws/ecs/test/main',
    'awslogs-region': 'ap-northeast-1',
    'awslogs-stream-prefix': 'ecs',
  },
  logDriver: 'awslogs',
},
```

This bug was the result of not handling a `TypeReference::Map` structure type inside the `ResourceIr::Object` value type while emitting resources for each synthesizer. Instead, like as structure of `TypeReference::Primitive`, the entry key should not have its case converted.

Note: This was unimplemented for java, golang, and csharp, but this PR implements this for all three and fixes the bug for python and typescript.

Fixes #[29027](https://github.com/aws/aws-cdk/issues/29027)


TheRealAmazonKendra (changes requested):
So, I don't think this is quite the right fix. It shouldn't matter what the name is when we are handling a named type. What I think is actually happening is that the named type `LogConfiguration` contains a field `Options` of type `json`. The code here doesn't recurse enough to discover that.

TheRealAmazonKendra (approved):

PR #594, opened 2024-03-05
chore(ir): add specific error message when an invalid resource property is used

colifran:
In the event that an invalid property is specified for a specific resource type the `value_type` will be `None` causing the `other` arm of the match statement below to execute.

https://github.com/cdklabs/cdk-from-cfn/blob/2c14390f73b765b961a71b0cc2952fe87af587fb/src/ir/resources/mod.rs#L119-L129

Due to the `unimplemented!` macro panicking in-place, a non-descriptive error message is displayed while using cdk migrate:

```
cdk migrate --from-path './CDKMigrateExampleTemplate.yml' --stack-name CustomStackName
...
 ❌  Migrate failed for `CustomStackName`: stack generation failed due to error 'unreachable'

stack generation failed due to error 'unreachable'
```

This PR adds a new explicit check for the case where `property_type` is `None`. If that case is true, then a more descriptive error message will be displayed explaining what resource property is not valid. For example,

```
cdk migrate --from-path './CDKMigrateExampleTemplate.yml' --stack-name CustomStackName
...
 ❌  Migrate failed for `CustomStackName`: stack generation failed due to error 'ReadEndpoint is not a valid resource property for resource RDSCluster of type AWS::RDS::DBCluster'

stack generation failed due to error 'ReadEndpoint is not a valid resource property for resource RDSCluster of type AWS::RDS::DBCluster'
```

### DIFF
--- a/src/ir/resources/mod.rs
+++ b/src/ir/resources/mod.rs
@@ -402,7 +402,7 @@ impl ResourceInstruction {
                     )
                     .replace('\"', "");
                     return Err(TransmuteError::new(format!(
-                        "{name} is not a valid property type for resource type {resource_type}"
+                        "{name} is not a valid property for resource type {resource_type}"
                     )));
                 }
                 let translator = ResourceTranslator {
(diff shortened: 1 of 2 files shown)


TheRealAmazonKendra:
Could we wordsmith that error a little more? Maybe some thing more like: `${stackName} could not be generated because <the rest of your error message is great>`

TheRealAmazonKendra:
Neverminddddd, that part of the error message is on the CLI side so approving this since it looks great!

TheRealAmazonKendra (approved):

PR #621, opened 2024-03-19
fix(ir): implement support for `ResourceValue::Object`s containing a nested `TypeReference::Union`s in deserialized templates

### LINKED ISSUE #624: bug: resource translation fails when a `ResourceValue::Object` containing a nested `TypeReference::Union` is encountered in deserialized template
When a deserialized input template has a property that is a `ResourceValue::Object` containing a nested `TypeReference::Union`, the resource translation will fail because it is currently unimplemented - https://github.com/cdklabs/cdk-from-cfn/blob/fd3ec9e73278d9cc9b476d43140cfb077dad3963/src/ir/resources/mod.rs#L133.

As an example, consider the following template:
```
ClientFunction:
    Type: AWS::Serverless::Function
    Properties:
      FunctionName: 'pricing-client'
      Handler: client_lambda.lambda_handler
      MemorySize: 512
      Timeout: 300
      CodeUri: ./pricing-client/
      PermissionsBoundary: !Sub 'arn:aws:iam::${AWS::AccountId}:policy/GithubActionsIamResourcePermissionsBoundary'
      Role: !GetAtt LambdaExecutionRole.Arn
      Runtime: python3.11
      Events:
        ApiEvent:
          Type: Api
          Properties:
            Path: /path
            Method: get
```

The `Events` property is a event source that must contain one of the following event types: AlexaSkill, Api, CloudWatchEvent, CloudWatchLogs, Cognito, DocumentDB, DynamoDB, EventBridgeRule, HttpApi, IoTRule, Kinesis, MQ, MSK, S3, Schedule, ScheduleV2, SelfManagedKafka, SNS, SQS. As a result, the corresponding type is a `TypeReference::Union(TypeUnion::Static([TypeReference::Named(_), ..]))` which translates to meaning that the event type value must be one of the event types listed above.

We should add support for the `TypeReference::Union` type. This could be supported via a `UnionOf` struct similar to the `MapOf` struct used for the `TypeReference::Map` type.

colifran:
Fixes #624


TheRealAmazonKendra:
Can you help me understand the issue that you're fixing here?

PR #644, opened 2024-04-04
chore: add end-to-end test for ApiGatewayV2

SankyRed:


tests/end-to-end.rs
```diff
@@ -84,6 +84,7 @@ test_case!(cloudwatch, "CloudwatchStack", &["golang"]);
 test_case!(ecs, "EcsStack", &["java", "golang"]);
 test_case!(ec2, "Ec2Stack", &["java", "golang"]);
 test_case!(efs, "EfsStack", &["java", "golang"]);
+test_case!(apigatewayv2, "ApiGatewayV2Stack", &["csharp", "golang", "python"]);
```
colifran, line 87: I'm concerned about not synthesizing python and csharp. If these aren't working you may have uncovered a bug. Could you remove these and we can debug the failing tests?

colifran (changes requested):
Looks like this is going in the right direction! See my comment below.

PR #661, opened 2024-04-29
fix(all): a single allowed value causes failure

TheRealAmazonKendra:
We are checking for the second value of allowed values first when checking for booleans. This causes a failure when only one allowed value is provided for a parameter.

This PR includes the template that is being tested against but not the test cases because there are still other failures. The next PR will include the test cases and a fix for the bug preventing a successful run.

Fixes part of https://github.com/aws/aws-cdk/issues/29752


colifran (approved):
LGTM!

PR #751, opened 2024-11-25
fix: allow translation of doubles from strings

### LINKED ISSUE #750: string to float translation does not work
I noticed recently when creating some cloudwatch alarms that if you input a string with a float/double, the program crashes (it also doesn't really tell you WHERE it crashes, which is saddening).

When looking deeper into the code, this failure occurs when the resolving type in IR uses specification data, and sees a Number, and tries to parse a String -> Int, but in this case the string is a float (also a valid number).

From the integration tests example, note **Threshold**
```
    "MyApi5xxErrorsAlarm": {
      "Type": "AWS::CloudWatch::Alarm",
      "Properties": {
        "AlarmDescription": "Example alarm",
        "Namespace": "AWS/ApiGateway",
        "Dimensions": [
          {
            "Name": "ApiName",
            "Value": "MyApi"
          }
        ],
        "MetricName": "5XXError",
        "ComparisonOperator": "GreaterThanThreshold",
        "Statistic": "Average",
        "Threshold": "0.005",
        "Period": 900,
        "EvaluationPeriods": 1,
        "TreatMissingData": "notBreaching",
        "AlarmActions": [
          {
            "Fn::ImportValue": {
              "Fn::Sub": [
                "${Environment}AlarmsTopicArn",
                {
                  "Environment": {
                    "Ref": "EnvironmentName"
                  }
                }
              ]
            }
          }
        ]
      }
    }
```

iph:
There were certain templates that errored out and said "not a valid integer". When looking at the code, it turns out that if a string thinks there is a number -- a number in JSON can be either integer or doable. Originally, the code parsed only ints, so any float/double fails outright. This update now will try both. I'm sure in the future we will need to try to parse one, fail, parse the other -- but for now, just look for a ".".

Fixes #750

### DIFF
--- a/src/ir/resources/mod.rs
+++ b/src/ir/resources/mod.rs
@@ -85,11 +85,20 @@ impl<'a, 'b> ResourceTranslator<'a, 'b> {
                             }
                         })?)),
                         Primitive::Number => {
-                            Ok(ResourceIr::Number(s.parse().map_err(|cause| {
-                                Error::ResourceTranslationError {
-                                    message: format!("{cause}"),
-                                }
-                            })?))
+                            let ir = match s.contains(".") {
+                                true => ResourceIr::Double(WrapperF64::new(s.parse().map_err(
+                                    |cause| Error::ResourceTranslationError {
+                                        message: format!("{cause}"),
+                                    },
+                                )?)),
+                                false => ResourceIr::Number(s.parse().map_err(|cause| {
+                                    Error::ResourceTranslationError {
+                                        message: format!("{cause}"),
+                                    }
+                                })?),
+                            };
+
+                            Ok(ir)
                         }
                         _ => Ok(ResourceIr::String(s)),
                     };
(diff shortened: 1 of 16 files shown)


mrgrain:
> Seems failure is from:
> 
> https://github.com/cdklabs/cdk-from-cfn/blob/main/.github/workflows/status-checks.yml#L81-L82
> 
> Did we add OIDC enablement?

I think you are right. They have release a new major version recently which had a lot of bugs at release. Can you try reverting the action back to v4 and see if that does it?

iph:
Agh. One day codecov, one day. Yeah, let me revert codecov to v4.

mrgrain:
Otherwise approved.

iph:
nice. Will first get out the revert, then will rerun workflows, just to get it flowing again.

iph:
no worries fam, love codecov despite the minor hiccups in v5 action :D

PR #1003, opened 2025-08-21
fix: number parameters are converted to strings

### LINKED ISSUE #974: Bug: Wrong typescript types generated for template parameters other than string
For a CFN template with following parameters
```yaml
Parameters:
  env:
    Type: String
  kinesisStreamName:
    Type: String
  kinesisStreamShardCount:
    Type: Number
    Default: 1
```
the typescript type generated is
```ts
export interface MyStackName extends cdk.StackProps {
  /**
   */
  readonly env: string;
  /**
   */
  readonly kinesisStreamName: string;
  /**
   * @default 1
   */
  readonly kinesisStreamShardCount?: string;
}
```

Expectation: The tool should generate the correct parameter types.

TheRealAmazonKendra:
Fixes #974


src/synthesizer/csharp/mod.rs
```diff
@@ -164,14 +165,31 @@ impl Synthesizer for CSharp<'_> {
                         _ => "",
                     };
                     if let Some(v) = &param.default_value {
-                        cfn_param.line(format!(
-                            "Default = {list_optional_prefix}props.{name}{list_optional_suffix} ?? \"{}\",",
-                            v.escape_debug()
-                        ));
+                        let default_conversion = match param.constructor_type.as_str() {
+                            "Number" => {
+                                if param.no_echo.as_ref().is_some_and(|x| x == "true") {
```
rix0rrr, line 170: How does `no_echo` factor into this value being optional or not?

tests/end-to-end/batch/csharp/Stack.cs
```diff
@@ -24,6 +26,15 @@ public class BatchStack : Stack
 
         public BatchStack(Construct scope, string id, BatchStackProps props = null) : base(scope, id, props)
         {
+            // Applying default props
+            props ??= new BatchStackProps();
+            props.MaxCpus = new CfnParameter(this, "MaxCpus", new CfnParameterProps
+            {
+                Type = "Number",
+                Default = props.MaxCpus.ToString() ?? "64",
```
rix0rrr, line 34: `Default` doesn't need to be a string, it is `any`. It can also be a number.

tests/end-to-end/batch/csharp/Stack.diff
```diff
@@ -1,13 +1,21 @@
 diff --git a/./tests/end-to-end/batch/template.json b/tests/end-to-end/batch-csharp-working-dir/cdk.out/Stack.template.json
-index c503353..684c8cb 100644
+index d75fe65..afb2161 100644
 --- a/./tests/end-to-end/batch/template.json
 +++ b/tests/end-to-end/batch-csharp-working-dir/cdk.out/Stack.template.json
-@@ -1,62 +1,5 @@
+@@ -1,69 +1,12 @@
```
rix0rrr, line 5: Why is did this diff change so much? Can we sort the JSON output first (in a separate PR) so that the diff becomes legible?

Issue #509, opened 2024-01-16
End-to-end tests: migrate away from using deprecated APIs in the AWS SDK for Rust


madeline-k:
Now the the AWS SDK for Rust is GA, some APIs in the dev preview version have been deprecated. These APIs continue to work fine, but we should migrate away from using them.

Issue #507, opened 2024-01-16
End-to-end tests: Use a separate file for defining test cases


madeline-k:
Currently all of the test logic, and each test case is in the same file `end-to-end.rs`. It would reduce cognitive load when adding new tests, if that was  in a separate file containing only the test cases, and not all of the logic for executing them.

Issue #506, opened 2024-01-16
End-to-end tests: Use a better mechanism to compare CloudFormation templates


madeline-k:
The end-to-end tests currently use `git diff` to compare input and output CloudFormation templates. This is not sophisticated enough, because files of two templates may look different, even when the CloudFormation templates are equivalent. 

A couple options to implement a more advance mechanism here:
1. Do JSON object comparison. We should also consider automatically resolving differences where the JSON is different, but the CloudFormation templates are equivalent.
2. Create a CloudFormation stack from the original template, then do a `cdk diff` with the new stack. This will have challenges with handling credentials if we want to do it in a pipeline or in GitHub actions.

PR #1031, opened 2025-09-09
testing: improve end-to-end testing

TheRealAmazonKendra:
With this change, tests pass or synth is skipped with a link to an open GitHub issue.

What this PR does not do: Any bug fixes in src. With a change this size, it is too risky to mix both code fixes and fixes to the testing infrastructure in the same PR. It would be very difficult to identify any new bugs potentially introduced. To avoid regressions, this PR focuses solely on fixing the testing mechanisms and provides open issue links for all the issues causing test skips of any kind. These fixes will be forthcoming in future PRs.

This change builds off of https://github.com/cdklabs/cdk-from-cfn/pull/1005 with significant refactoring, code improvement, and a fully working set of test suites end to end. The amount of files changed here is a brutal so the commits were intentionally split in a way that hopefully makes it easier to review:
1. File moves and deletions: The general structure of the testing suite wasn't great so I updated it to be much more clear where everything is, when files are shared between the two synthesizer tests, and to reduce code duplication as much as possible.
2. Updates to test cases so the tests would run and all templates would be deployable*:  Some of the templates could not actually deploy and others could not synth because of a bug in the template, not in the language implementation. In this commit, all of these instances were fixed in this commit. (Upon review, one errant change made it into this commit: tests/cdk_stack_synth/synth/template.rs)
3. Updates to boilerplate files: I switched us to using a template for these files instead of writing them with a CodeBuffer. This simplifies things a bit and remove the dependency on CodeBuffer. 
4. The actual changes to the testing infrastructure. This is the "everything else" commit. It creates two new crates (that will not be published): `cdk-from-cfn-testing` and `cdk-from-cfn-testing-end-to-end`. 
    - `cdk-from-cfn-testing` contains all the logic for tests that are shared across the tests inside src and tests in the main crate. 
    - `cdk-from-cfn-testing-end-to-end` contains all the logic specific to the `end-to-end` feature. The end to end tests require a lot of AWS SDK dependencies that are large and sometimes conflict with wasm build. Separating this out improves the package size and build time significantly. Cargo also doesn't assess code for features that aren't currently active, moving the code into it's own crate makes it so the only lines that are ignored are where these functions are called and not where the implementation lives so issues are caught faster on build/test.

**For reviewing meaningful changes, I would suggest viewing the last three commits and filtering out the first**

Summary of changes in this PR:
1. Separation of tests for the IR Synthesizer (we absolutely overloaded this term) and the end-to-end-ish/end-to-end tests
    1. Our old end-to-end tests were only using internal logic to call the ir synthesizer an then manually write the template to the working dir. That internal logic resulted in us missing a bug during when using the actual dist. The internal synthesis logic would work and create the stack template, but the cdk-from-cfn command would error out on the same template. This PR moves a portion of the old tests (rename to ir_synthesizer_test) into the synthesizer folder since they are really technically unit tests for this scope. These tests only test for stack match between expected and actual.
    2. A second set of tests (cdk_stack_synth_test) with the same test cases are now in the tests folder to act as end-to-end tests. They use the built cdk-from-cfn distribution to generate the stack file, test for stack match between expected and actual, generate/copy app scaffolding files into the working dir, run `cdk synth` on the app, and then compare the two template json files against each other. If there is any difference, we write a Stack.diff file with acceptable differences that will not result in a diff when a change set is created between the two once deployed. Lastly, it compares all the created templates against each other. Because the comparison with the original template.json file gives leeway for some acceptable differences that we interpret the same, this is a safe guard. It does a literal template comparison and expects them to be exactly the same and does not leave room for human error in the interpretation of acceptable diff.
        - These tests are run in groups by the test name so the output only shows the test name, not the language. For an extremely detailed output of the steps taken for each test `-- --nocapture` can be used. 
   3. When the test is run with `--features end-to-end`, for each test will first deploy the original template and then will create a change set for each target language (synth skips excluded), the test fails if there are any changes in the change set. There is no acceptable diff here, so this step confirms that the acceptable diff really is acceptable.
        -  Unfortunately, cdk-from-cfn does not yet have any automation to run these from the repo so, for the moment, this still requires contributors to run these. 
 
2. Addition of a proc macro to build both sets of tests from one list of test cases. This ensures one source of truth.
3. The addition of `cdk-from-cfn-testing` and `cdk-from-cfn-testing-end-to-end` to avoid circular dependencies, dependency issues impacting wasm build, etc.
6. A central handling of files, paths, etc. As mentioned above, cdk-from-cfn used a lot of paths that were written in plain text. This was making it extremely difficult to make changes to the tests without breaking something. The error handling was also super unclear about what exactly broke in these circumstances. So, Paths builds all paths used with as little plain text as possible in these functions and Files uses paths to perform all read and write operations. This is perhaps overkill, but it saved a lot of troubleshooting time.
7. The use of features instead of env variables. `end-to-end`, `skip-clean`, `update-snapshots` and `pre-install` were all added. 
      - `end-to-end` runs the end to end tests
      - `skip-clean` skips the cleanup of any tests you run. If you are running end-to-end tests, the stacks also remain
      - `update-snapshots` updates the expected stack files, and the acceptable diff if there are any differences. No tests will fail in diff when this flag is used
      - `pre-install` downloads all the dependencies for the cdk-stack-synth tests at build time. This is enabled by default but can be disabled if you do not have access to the internet but still want to run the synthesizer tests or other unit tests. Pre-installing saves a significant amount of time while running the cdk-stack-synth tests.
9. Extensive readme updates: one set of changes is big enough that it isn't open by default. I'm particularly interested in feedback on whether these are clear enough.

*GroundStation is not deployable and will never be due to permissions but a [bug](https://github.com/aws/aws-cdk/issues/29752) was reported to us a long time ago using this template as an example of a bug to synth. It is a useful test case for synth only for that particular issue. 

Closes https://github.com/cdklabs/cdk-from-cfn/issues/509, https://github.com/cdklabs/cdk-from-cfn/issues/506, https://github.com/cdklabs/cdk-from-cfn/issues/507


rix0rrr (approved):
I can't review this. I'll trust you instead 😉

PR #1138, opened 2026-01-30
feat: add CloudFormation custom resource (custom::*) support

sai-ray:
### Description:

`cdk-from-cfn` currently fails when encountering `Custom::*` and `AWS::CloudFormation::CustomResource` resources because they aren't in the CDK schema. In the IR translation phase, every property is validated against the schema, and since custom resources have no schema entry, all properties (including `ServiceToken`) fail validation.

This PR adds support for converting both `Custom::*` and `AWS::CloudFormation::CustomResource` resources to CDK code across all five target languages (TypeScript, Python, Go, Java, C#).

The implementation relies on the L1 `CfnCustomResource` construct with escape hatches, consistent with how the codebase handles all other resources via L1 `Cfn*` constructs:

- `CfnCustomResource` constructor takes `serviceToken` (the only typed property)
- `addOverride('Type', 'Custom::XXX')` sets the actual CloudFormation type (skipped for AWS::CloudFormation::CustomResource since the type is already correct)
- `addPropertyOverride()` passes through all custom properties with their original names
- DeletionPolicy, Metadata, UpdatePolicy, DependsOn, and Condition all use the same L1 patterns as standard resources

`AWS::CloudFormation::CustomResource` is normalized into the same `Custom` pipeline during parsing via a `CFN_CUSTOM_RESOURCE` constant, so both resource types share the same code path with a single conditional to skip the type override.

### Example:

`Custom::*` Input: 
```TypeScript
"MyCustomResource": {
  "Type": "Custom::DatabaseSetup",
  "DeletionPolicy": "Retain",
  "DependsOn": ["BackingLambda"],
  "Properties": {
    "ServiceToken": { "Fn::GetAtt": ["BackingLambda", "Arn"] },
    "DatabaseName": "mydb",
    "TableCount": 5,
    "EnableLogging": "true",
    "Tags": ["prod", "critical"]
  }
}
```

Output:

Before (fails with error):
```TypeScript
ServiceToken is not a valid property for resource MyCustomResource of type Custom::DatabaseSetup
```

After:
```TypeScript
const myCustomResource = new cdk.CfnCustomResource(this, 'MyCustomResource', {
  serviceToken: backingLambda.attrArn,
});
myCustomResource.addOverride('Type', 'Custom::DatabaseSetup');
myCustomResource.addPropertyOverride('DatabaseName', 'mydb');
myCustomResource.addPropertyOverride('TableCount', 5);
myCustomResource.addPropertyOverride('EnableLogging', 'true');
myCustomResource.addPropertyOverride('Tags', ['prod', 'critical']);
myCustomResource.cfnOptions.deletionPolicy = cdk.CfnDeletionPolicy.RETAIN;
myCustomResource.addDependency(backingLambda);
```

`AWS::CloudFormation::CustomResource` Input: 
```TypeScript
"CfnCustomResource": {
  "Type": "AWS::CloudFormation::CustomResource",
  "Properties": {
    "ServiceToken": { "Fn::GetAtt": ["BackingLambda", "Arn"] },
    "Region": "us-west-2"
  }
}
```

Output:

Before (fails with error):
```TypeScript
Region is not a valid property for resource CfnCustomResource of type AWS::CloudFormation::CustomResource
```

After:
```TypeScript
const cfnCustomResource = new cdk.CfnCustomResource(this, 'CfnCustomResource', {
  serviceToken: backingLambda.attrArn,
});
// No addOverride('Type', ...) - type is already correct
cfnCustomResource.addPropertyOverride('Region', 'us-west-2');
```

GetAtt references to custom resource outputs:
```TypeScript
environment: {
  variables: {
    'DB_ENDPOINT': myCustomResource.getAtt('Endpoint').toString(),
    'CFN_RESULT': cfnCustomResource.getAtt('Result').toString(),
  },
},
```

### Changes:

Intermediate Representation (IR) layer `(src/ir/)`:

- ServiceToken validation: Custom resources without `ServiceToken` fail early with a clear error, rather than producing broken CDK output that fails at synth time.
- JSON passthrough: All custom properties bypass schema validation and are treated as `Json` type, because custom resources have no schema and their properties are arbitrary key-value pairs passed directly to the backing Lambda.
- `is_custom_resource` flag on `Origin::GetAttribute`: Synthesizers need to know whether a `GetAtt` target is a custom resource so they can emit dynamic `getAtt('X')` calls instead of typed `attrX` accessors, since `CfnCustomResource` has no typed attribute properties.
- Import generation skips `Custom::*` and `AWS::CloudFormation::CustomResource` resources since `CfnCustomResource` comes from `aws-cdk-lib` core which is always imported as `cdk`, so no additional service-level import is needed.

Synthesizers `(src/synthesizer/)`:

- New `emit_custom_resource` function in each of the 5 language synthesizers, called via early return from `emit_resource`. This keeps the custom resource code path completely separate so existing standard resource logic is untouched and unaffected.
- Handles conditional resources, DeletionPolicy, Metadata, UpdatePolicy, and DependsOn using the same L1 patterns as standard resources, so the generated code is consistent across resource types.
- GetAtt references to custom resources use dynamic lookup (`.getAtt('X').toString()`) in the absence of typed accessors (`.attrX`). The `.toString()` is required because `getAtt()` returns a `Reference` object, and without it, strictly-typed contexts like `Map<String, String>` in Java or `Dictionary<string, string>` in C# would fail to compile.
- Type override `(addOverride('Type', 'Custom::XXX'))` is skipped for `AWS::CloudFormation::CustomResource` since the type is already correct.

Tests:

- Added snapshot tests with a template that mixes standard and custom resources, covering DeletionPolicy, DependsOn, and GetAtt cross-references. Expected outputs added for all 5 languages in both Stack and Construct modes.
- Added unit tests in each synthesizer covering basic emit, conditional branching, GetAtt dynamic lookup, metadata, update policy, and deletion policy variants.
- Added IR-level unit tests for ServiceToken validation, JSON passthrough, GetAtt flag propagation and the standard resource invalid property error path.

### Task List

- [x]  Add is_custom_resource field to Origin::GetAttribute
- [x]  Update all synthesizers to use getAtt() for custom resource attribute references
- [x]  Add ServiceToken validation
- [x]  Treat custom resource properties as JSON passthrough
- [x]  Skip CDK import generation for custom resources
- [x]  Add custom resource detection and CustomResource construct generation
    - [x]   Add emit_custom_resource() for TypeScript
    - [x]  Add emit_custom_resource() for Python
    - [x]  Add emit_custom_resource() for Go
    - [x]  Add emit_custom_resource() for Java
    - [x]  Add emit_custom_resource() for C#
- [x] Add and update tests
  - [x] Add snapshot cases for custom resources (all 5 languages)
  - [x] Update integration test pipeline (skip list, macro registration)
  - [x] Add unit tests for TypeScript synthesizer
  - [x] Add unit tests for Python synthesizer
  - [x] Add unit tests for Go synthesizer
  - [x] Add unit tests for Java synthesizer
  - [x] Add unit tests for C# synthesizer
  - [x] Add unit tests for IR resources (ServiceToken validation, JSON passthrough, GetAtt flag)
- [x] Add support for `AWS::CloudFormation::CustomResource` resource type.
  - [x] Normalize `AWS::CloudFormation::CustomResource` into the Custom pipeline
  - [x] Skip type override for `AWS::CloudFormation::CustomResource`
  - [x] Add unit tests for `AWS::CloudFormation::CustomResource `
  - [x] Update snapshot cases 

Related to https://github.com/aws-amplify/amplify-cli/issues/14539

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -639,9 +639,9 @@ checksum = "46c5e41b57b8bba42a04676d81cb89e9ee8e859a1a66f80a5a72e1cb76b34d43"
 
 [[package]]
 name = "bytes"
-version = "1.10.1"
+version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d71b6127be86fdcfddb610f7182ac57211d4b18a3e9c82eb2d17662f2227ad6a"
+checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
 
 [[package]]
 name = "bytes-utils"
@@ -676,7 +676,7 @@ dependencies = [
 
 [[package]]
 name = "cdk-from-cfn"
-version = "0.273.0"
+version = "0.285.0"
 dependencies = [
  "base64",
  "cdk-from-cfn-macros",
@@ -791,18 +791,18 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.56"
+version = "4.5.60"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a75ca66430e33a14957acc24c5077b503e7d374151b2b4b3a10c83b4ceb4be0e"
+checksum = "2797f34da339ce31042b27d23607e051786132987f595b02ba4f6a6dffb7030a"
 dependencies = [
  "clap_builder",
 ]
 
 [[package]]
 name = "clap_builder"
-version = "4.5.56"
+version = "4.5.60"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "793207c7fa6300a0608d1080b858e5fdbe713cdc1c8db9fb17777d8a13e63df0"
+checksum = "24a241312cea5059b13574bb9b3861cabf758b879c15190b37b6d6fd63ab6876"
 dependencies = [
  "anstream",
  "anstyle",
@@ -812,9 +812,9 @@ dependencies = [
 
 [[package]]
 name = "clap_lex"
-version = "0.7.5"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b94f61472cee1439c0b966b47e3aca9ae07e45d070759512cd390ea2bebc6675"
+checksum = "3a822ea5bc7590f9d40f1ba12c0dc3c2760f3482c6984db1573ad11031420831"
 
 [[package]]
 name = "cmake"
@@ -1198,9 +1198,9 @@ checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
 
 [[package]]
 name = "futures"
-version = "0.3.31"
+version = "0.3.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "65bc07b1a8bc7c85c5f2e110c476c7389b4554ba72af57d8445ea63a576b0876"
+checksum = "8b147ee9d1f6d097cef9ce628cd2ee62288d963e16fb287bd9286455b241382d"
 dependencies = [
  "futures-channel",
  "futures-core",
@@ -1213,9 +1213,9 @@ dependencies = [
 
 [[package]]
 name = "futures-channel"
-version = "0.3.31"
+version = "0.3.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2dff15bf788c671c1934e366d07e30c1814a8ef514e1af724a602e8a2fbe1b10"
+checksum = "07bbe89c50d7a535e539b8c17bc0b49bdb77747034daa8087407d655f3f7cc1d"
 dependencies = [
  "futures-core",
  "futures-sink",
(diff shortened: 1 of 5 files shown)


src/ir/importer/mod.rs
```diff
@@ -24,6 +24,11 @@ impl ImportInstruction {
         for (_, resource) in parse_tree {
             let type_name = &resource.resource_type;
 
+            // Skip Custom resources - they don't have CDK service modules
+            if type_name.starts_with("Custom::") {
+                continue;
```
sai-ray: Custom resources don't map to any CDK service module. They use `CfnCustomResource` from `aws-cdk-lib` core, which is always available. So we skip them here to avoid a failed import lookup.

src/synthesizer/csharp/mod.rs
```diff
@@ -359,6 +363,87 @@ impl Synthesizer for CSharp<'_> {
...
+        trailing_newline: true,
+    });
+    if let Some(token) = service_token {
+        resource_constructor.text("ServiceToken = ");
+        token.emit_csharp(&resource_constructor, schema, class_type)?;
+        resource_constructor.text(",");
+        resource_constructor.newline();
+    }
+
+    // Override the type from AWS::CloudFormation::CustomResource to Custom::XXX
```
otaviomacedo: Why?

cdk-from-cfn-testing/cases/custom_resource/template.json
```diff
@@ -0,0 +1,87 @@
...
+                        "Arn"
+                    ]
+                },
+                "Code": {
+                    "ZipFile": "def handler(event, context):\n  return {'Status': 'SUCCESS', 'Data': {'Endpoint': 'test-endpoint'}}\n"
+                }
+            }
+        },
+        "MyCustomResource": {
+            "Type": "Custom::DatabaseSetup",
```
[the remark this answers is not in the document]
sai-ray: Added. Both `Custom::*` and `AWS::CloudFormation::CustomResource` are now supported and tested (unit tests + snapshot tests).

Issue #694, opened 2024-06-13
Named conditions fail to migrate


steven10172:
When attempting to migrate a CloudFormation template that contains a named condition it failed with an error.

CloudFormation:
```
Conditions:
  HasSecurityGroupIds:
    Fn::Not:
      - Fn::Equals:
        - ""
        - Fn::Join: ["", Ref: SecurityGroupIds]


  CloudAuthVpcEndpoint:
    Type: AWS::EC2::VPCEndpoint
    Properties:
      SecurityGroupIds:
        Fn::If:
          - HasSecurityGroupIds
          - Ref: SecurityGroupIds
          - Ref: 'AWS::NoValue'
      ServiceName:
        Fn::FindInMap:
          - CloudAuthVpcEndpointServices
          - Ref: 'AWS::Region'
          - vpces
      SubnetIds:
        - Fn::ImportValue: PrivateSubnet01
        - Fn::ImportValue: PrivateSubnet02
        - Fn::ImportValue: PrivateSubnet03
      VpcEndpointType: Interface
      VpcId:
        Fn::ImportValue: VPC
```

Error:
```
CloudAuthInfrastructureStackStack could not be generated because Conditions.HasSecurityGroupIds: data did not match any variant of untagged enum Singleton at line 22 column 5
```

iph:
The error is weird but shockingly isn't named conditions that is the problem. It seems that Fn::Equals in combination with Fn::Join is `not a great combo`.

I'm unsure why. As an example, I took your template and modified to remove the Fn::Join:

```
Conditions:
  HasSecurityGroupIds:
    Fn::Not:
      - Fn::Equals:
        - ""
        - Ref: AWS::Region
Resources:
  CloudAuthVpcEndpoint:
    Type: AWS::EC2::VPCEndpoint
    Properties:
      SecurityGroupIds:
        Fn::If:
          - HasSecurityGroupIds
          - Ref: SecurityGroupIds
          - Ref: 'AWS::NoValue'
      ServiceName:
        Fn::FindInMap:
          - CloudAuthVpcEndpointServices
          - Ref: 'AWS::Region'
          - vpces
      SubnetIds:
        - Fn::ImportValue: PrivateSubnet01
        - Fn::ImportValue: PrivateSubnet02
        - Fn::ImportValue: PrivateSubnet03
      VpcEndpointType: Interface
      VpcId:
        Fn::ImportValue: VPC
```

and gets past the issue.

[...]

PR #1234, opened 2026-06-15
fix: deserialize shorthand-tagged intrinsics in Fn::Not and Fn::Sub

darwinboersma:
Refs #694, aws/aws-cdk-cli#824

## Problem

Migrating CloudFormation templates that use YAML shorthand intrinsics in two common places fails with:

```
... could not be generated because <path>: untagged and internally tagged enums do not support enum input
```

Two spots in the parser model their payloads as `#[serde(untagged)]` enums:

- `Singleton` — the operand of `Fn::Not`
- `SubPayload` — the `Fn::Sub` `[template, {variables}]` form

serde's untagged deserialization buffers the input through `Content` and then refuses *enum* input, but `serde_yaml` surfaces YAML shorthand tags (`!Equals`, `!Ref`, `!GetAtt`, …) as enum input. So a shorthand intrinsic nested under `Fn::Not`, or inside a `Fn::Sub` variables map, fails to parse — e.g.:

```yaml
# condition operand
!Not [!Equals [!Ref Foo, ""]]
# Sub variables map
!Sub ["${Foo}", { Foo: !Ref Bar }]
```

## Fix

Replace both untagged enums with hand-written `Deserialize` impls that deserialize the inner `ConditionValue` / `ResourceValue` directly. Those types already have visitors that understand YAML tags, so deserializing them directly avoids the `Content` buffering that rejected the tagged input. Behavior for the previously-working forms (bare value, single-element list, `[template, {literals}]`) is preserved.

This is parser-only and language-agnostic — no synthesizer, IR, or schema changes.

## Scope

This fixes the *deserialization* failure. It does not change which intrinsic functions are permitted *inside* a `Conditions` block — e.g. `Fn::Join`/`Fn::Sub` used inside a condition is a separate gap. That's why this references #694 rather than closing it.

## Testing

- Added regression tests `function_not_nested_shorthand` (conditions) and `intrinsic_sub_shorthand_tagged_variable` (resources).
- `cargo test --all-features` — all tests pass.
- `cargo fmt --check` and `cargo clippy --tests -- -Adead-code -Dwarnings -Dclippy::dbg_macro` — clean.

### DIFF
--- a/src/parser/condition/mod.rs
+++ b/src/parser/condition/mod.rs
@@ -249,19 +249,84 @@ impl<'de> serde::Deserialize<'de> for ConditionValue {
     }
 }
 
-#[derive(serde::Deserialize)]
-#[serde(untagged)]
-enum Singleton {
-    Value(ConditionValue),
-    SingletonTuple((ConditionValue,)),
-}
+// `Fn::Not` takes its operand either as a bare condition value or wrapped in a
+// single-element list. This was previously modeled with `#[serde(untagged)]`,
+// but serde's untagged deserialization buffers the input through `Content` and
+// then refuses enum input ("untagged and internally tagged enums do not support
+// enum input"). serde_yaml surfaces shorthand tags such as `!Equals` AS enum
+// input, so `!Not [!Equals [...]]` failed to parse. Deserializing the inner
+// `ConditionValue` directly — it has its own visitor that understands tags —
+// avoids the buffering entirely.
+struct Singleton(ConditionValue);
 
 impl Singleton {
     fn unwrap(self) -> ConditionValue {
-        match self {
-            Self::Value(value) => value,
-            Self::SingletonTuple((value,)) => value,
+        self.0
+    }
+}
+
+impl<'de> serde::Deserialize<'de> for Singleton {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        struct SingletonVisitor;
+        impl<'de> serde::de::Visitor<'de> for SingletonVisitor {
+            type Value = ConditionValue;
+
+            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+                formatter.write_str("a condition value or a single-element list containing one")
+            }
+
+            fn visit_seq<A: serde::de::SeqAccess<'de>>(
+                self,
+                mut seq: A,
+            ) -> Result<Self::Value, A::Error> {
+                let value: ConditionValue = match seq.next_element()? {
+                    Some(value) => value,
+                    None => return Err(A::Error::invalid_length(0, &self)),
+                };
+                // `Fn::Not` takes exactly one operand; drain any extras.
+                while seq.next_element::<serde::de::IgnoredAny>()?.is_some() {}
+                Ok(value)
+            }
+
+            // Bare (non-list) scalar forms mirror ConditionValue's own handling.
+            fn visit_str<E: Error>(self, val: &str) -> Result<Self::Value, E> {
+                Ok(ConditionValue::String(val.into()))
+            }
+            fn visit_bool<E: Error>(self, val: bool) -> Result<Self::Value, E> {
+                Ok(ConditionValue::String(val.to_string()))
+            }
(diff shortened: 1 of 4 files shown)


src/parser/condition/mod.rs
```diff
@@ -249,19 +249,84 @@ impl<'de> serde::Deserialize<'de> for ConditionValue {
...
+            }
+            fn visit_i64<E: Error>(self, val: i64) -> Result<Self::Value, E> {
+                Ok(ConditionValue::String(val.to_string()))
+            }
+            fn visit_u64<E: Error>(self, val: u64) -> Result<Self::Value, E> {
+                Ok(ConditionValue::String(val.to_string()))
+            }
+            fn visit_f64<E: Error>(self, val: f64) -> Result<Self::Value, E> {
+                Ok(ConditionValue::String(val.to_string()))
+            }
```
mrgrain: These arms are a second copy of how `ConditionValue` already handles scalars, so the two can quietly drift apart over time. Can we delegate to `ConditionValue` here the way the `visit_map`/`visit_enum` arms just below do, instead of rebuilding the strings by hand? If there's no clean way to delegate the owned-scalar cases, let's at least add tests that hit each of these arms (bare bool/int/float/string operands to `!Not`) so we'd catch a divergence.

src/parser/condition/mod.rs
```diff
@@ -249,19 +249,84 @@ impl<'de> serde::Deserialize<'de> for ConditionValue {
...
+            fn visit_seq<A: serde::de::SeqAccess<'de>>(
+                self,
+                mut seq: A,
+            ) -> Result<Self::Value, A::Error> {
+                let value: ConditionValue = match seq.next_element()? {
+                    Some(value) => value,
+                    None => return Err(A::Error::invalid_length(0, &self)),
+                };
+                // `Fn::Not` takes exactly one operand; drain any extras.
+                while seq.next_element::<serde::de::IgnoredAny>()?.is_some() {}
```
mrgrain: Let's not silently swallow extra operands here. `Fn::Not` takes exactly one, so `!Not [a, b]` is a broken template and quietly dropping `b` hides the mistake from the user. Please return an error when there's a trailing element instead of draining it.

src/parser/intrinsics/mod.rs
```diff
@@ -214,20 +214,52 @@ impl StringOrPair {
...
+                self,
+                mut seq: A,
+            ) -> Result<Self::Value, A::Error> {
+                let template: String = match seq.next_element()? {
+                    Some(template) => template,
+                    None => return Err(A::Error::invalid_length(0, &self)),
+                };
+                // The second element (the variables map) is optional.
+                let variables: Option<ResourceValue> = seq.next_element()?;
+                while seq.next_element::<serde::de::IgnoredAny>()?.is_some() {}
```
mrgrain: Same as the `Fn::Not` case: a `Fn::Sub` with 3+ elements is malformed, and draining the extras hides that. Let's reject the trailing elements with an error rather than ignoring them.

mrgrain (commented):
A couple of inline suggestions on the new visitors.

mrgrain (changes requested):
Nice and thank you! Please increase test coverage for the changed files to meet threshold and resolve my requests. This should be mergeable soon.

src/parser/condition/mod.rs
darwinboersma: Good call, sharing that handling of scalars seems ideal and saves some boilerplate! I'll dig into that now

darwinboersma:
Thanks for the review! Addressed all three in acf7cba:

1. **Scalar arms duplicating `ConditionValue`** — the non-list `Fn::Not` operand now delegates to `ConditionValue` for every form (scalars route through `serde::de::value` deserializers, the same way the `visit_map`/`visit_enum` arms already did), so there's a single source of truth and no risk of drift.
2. **`Fn::Not` trailing operands** — now returns a deserialization error instead of draining them; `!Not [a, b]` (and `!Not []`) are rejected.
3. **`Fn::Sub` trailing elements** — same treatment; a 3rd element is now an error rather than silently dropped.

Also added tests covering the delegated scalar/map/enum arms and the new empty-list / extra-element error paths, which should bring patch coverage back over the threshold.

mrgrain (approved):

